Issue #18, opened 2016-08-19
Review name of main class: "BigchainDB": implict or explicit "DB"


sbellem:
The question is:

"Should we make the "DB" implicit, or explicit?"

```python
class Bigchain:
    """Main class to establish a connection to one or more BigchainDB nodes."""
```

OR

```python
class BigchainDB:
    """Main class to establish a connection to one or more BigchainDB nodes."""
```

ttmc:
The word "Bigchain" is a legacy word from the days when it was the name of the overall project. Yes, "BigchainDB" has many syllables, but it's the name of the project.

sohkai:
@ttmc But you almost never work with a `*DB` class when using a driver. You import RethinkDB as `rethink` (or just `r`); you import MongoDB as `mongo` in most instances. This pattern is pretty standard.

sohkai:
cc @diminator @vrde @TimDaub, would be nice to have some votes on this. (Just thumbs up or down on the issue is fine).

TimDaub:
> The word "Bigchain" is a legacy word from the days when it was the name of the overall project. Yes, "BigchainDB" has many syllables, but it's the name of the project.

I second that.

ttmc:
From the MongoDB docs:

"PyMongo is the recommended way to work with MongoDB from Python."

```python
import pymongo
```

sbellem:
@vrde @diminator @sohkai how about we close this issue?

sohkai:
Sure, I don't think we came to any consensus on this.

Issue #8, opened 2016-08-16
Define BichainDB API v1.1 Client API


sohkai:
Based on the API spec provided by https://github.com/bigchaindb/bigchaindb/issues/362, define the API that the Python client should provide. This should be generic enough such that a [Javascript](https://github.com/bigchaindb/bigchaindb/issues/256) / go / etc. client can be modelled after it.

May be useful to gain inspiration from:

* [elasticsearch-py](https://github.com/elastic/elasticsearch-py) (low-level client): [Honza Král - Lessons learned from building Elasticsearch client](https://youtu.be/XoZDdwz43uQ)
* [elastisearch-dsl-py](https://github.com/elastic/elasticsearch-dsl-py) (ORM-like client): [Honza Král - Designing a Pythonic Interface](https://youtu.be/FPQrUuddT0s)

sohkai:
Some thoughts from #21:
* The current actions should probably be suffixed with `transactions`
* The current actions could return an object that exposes a number of useful things to clients (e.g. `transaction.url`, `transaction.id`, `transaction.json`, `transaction.status` and etc.)

sbellem:
@sohkai  see #33

sohkai:
With #59 this should be done, barring any issues found during integration with pycoalaip and pycoalaip-bigchaindb.

PR #21, opened 2016-08-22
Arch

sbellem:


.travis.yml
```diff
@@ -13,7 +13,7 @@ before_install:
 
 install:
   - sudo apt-get install rethinkdb
-  - pip install "git+https://github.com/sbellem/bigchaindb.git@setup-no-pinning"
+  - pip install "git+https://github.com/bigchaindb/bigchaindb.git"
```
sohkai: Hmm, I guess this can point to the actual package on PyPI once we release a new version?

sbellem: Of course!

bigchaindb_driver/driver.py
```diff
@@ -39,6 +40,12 @@ def __init__(self, *, public_key, private_key, node):
         self.public_key = public_key
         self.private_key = private_key
         self.node = node
+        self.transport = transport_class(node)
+
+    def retrieve(self, txid):
+        path = '/transactions' + '/' + txid
```
sohkai: `/transactions/` should probably be in a api_urls.py module or something. Also, don't need the extra concat and (maybe) using `.format()` is better?

bigchaindb_driver/driver.py
```diff
@@ -84,8 +91,9 @@ def _push(self, tx):
         Return:
             The transaction pushed to the Federation.
         """
-        res = requests.post(self.node + '/transactions/', json=tx)
-        return res.json()
+        response = self.transport.forward_request(
+            method='POST', path='/transactions/', json=tx)
```
sohkai: See above.

bigchaindb_driver/driver.py
```diff
@@ -1,9 +1,9 @@
-import requests
-
 from bigchaindb_common import crypto
 from bigchaindb_common.transaction import Data, Fulfillment, Transaction
 from bigchaindb_common.exceptions import KeypairNotFoundException
 
+from .transport import Transport
+
 
 class BigchainDB:
```
sohkai: Rename to just `Bigchain`?

bigchaindb_driver/driver.py
```diff
@@ -39,6 +40,12 @@ def __init__(self, *, public_key, private_key, node):
         self.public_key = public_key
         self.private_key = private_key
         self.node = node
+        self.transport = transport_class(node)
+
+    def retrieve(self, txid):
```
sohkai: Should probably be renamed to `retrieve_transaction` or similar?

bigchaindb_driver/driver.py
```diff
@@ -39,6 +40,12 @@ def __init__(self, *, public_key, private_key, node):
...
+        path = '/transactions' + '/' + txid
+        response = self.transport.forward_request(method='GET', path=path)
+        return response.json()
```
sohkai: I wonder if it'd be useful to wrap the response back in a class, so that certain things are super easy for the user (e.g. `transaction.url`, `transaction.id`, `transaction.json`, `transaction.status` and etc.)

bigchaindb_driver/driver.py
sbellem: Yes. I thought of handling this  once we have decided on how to handle trailing slashes (bigchaindb/bigchaindb#585)

bigchaindb_driver/transport.py
```diff
@@ -0,0 +1,23 @@
+from .connection import Connection
+from .pool import Pool
```
sohkai: What's the status quo on relative imports? Are python3's relative imports OK to use?

bigchaindb_driver/transport.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Transport:
+
+    def __init__(self, *nodes):
+        self.nodes = nodes
+        self.init_pool(nodes)
+
+    def init_pool(self, nodes):
```
sohkai: Why not just use `self.nodes` instead of passing in a parameter?

bigchaindb_driver/transport.py
```diff
@@ -0,0 +1,23 @@
...
+        """Initialize the pool of connections."""
+        connections = [Connection(node_url=node) for node in nodes]
+        self.pool = Pool(connections)
+
+    def get_connection(self):
+        """Get a connection from the pool."""
+        return self.pool.get_connection()
+
+    def forward_request(self, method='GET', path='/', json=None):
```
sohkai: Could probably remove these default variables and just use `**kwargs` since `connection.request()` already has these defaults (or the other way around, but it probably makes more sense for these defaults to live in `Connection`).

bigchaindb_driver/driver.py
sbellem: tackled in #18

bigchaindb_driver/driver.py
sbellem: We can address this question as part #8

bigchaindb_driver/driver.py
sohkai: Can also be tackled in #8.

bigchaindb_driver/transport.py
sbellem: Good point, we can tackle this later as the implementation evolves.

bigchaindb_driver/transport.py
sbellem: Good point!

bigchaindb_driver/transport.py
sbellem: see https://www.python.org/dev/peps/pep-0328/

PR #31, opened 2016-08-23
Support multiple nodes for BigchainDB class

sbellem:


sohkai:
Might be nice to test the picker behaviour, at least in a unit test, but I'm not really concerned at this point.

sbellem:
@sohkai  yes, good point -- it gets tested now via [`tests.test_pool.test_get_connection`](https://github.com/sbellem/bigchaindb-driver/blob/ddd6853a5549d54e1a50308189e51bf556a370a0/tests/test_pool.py#L1-L9)

PR #40, opened 2016-08-24
Closes #26

### LINKED ISSUE #26: Expose ability to generate public/private keypair in driver
The driver should expose a way for the user to get a new public/private keypair for use with BigchainDB as they shouldn't have to care about any other BigchainDB package (e.g. bigchaindb-commons).

sbellem:


bigchaindb_driver/crypto.py
```diff
@@ -0,0 +1,19 @@
+from collections import namedtuple
+
+from cryptoconditions import crypto
+
+
+CryptoKeypair = namedtuple('CryptoKeypair', ('signing_key', 'verifying_key'))
```
sohkai: Should probably be `'CryptoKeyPair'`

PR #43, opened 2016-08-25
Development release 0.0.1dev1

sbellem:

PR #44, opened 2016-08-25
Change distribution format to not be a universal package

sohkai:
For now, we're not planning on supporting python2 so we shouldn't declare this package as universal.


sohkai:
Will be done in https://github.com/bigchaindb/bigchaindb-driver/pull/43

PR #61, opened 2016-08-29
Adjustments due to tx model change

### LINKED ISSUE #60: Adjust driver to slightly new interface of tx model
A few changes were made in the tx model of bigchaindb-common, hence the driver needs a little adjustment.

TimDaub:
Fixes #60.

These are some changes that had to be made as the transaction model changed slightly.

@sohkai this is mostly relevant to you at this point.

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -1,4 +1,4 @@
-from bigchaindb_common.transaction import Data, Fulfillment, Transaction
+from bigchaindb_common.transaction import Transaction
 
 from .crypto import generate_keypair
 from .exceptions import InvalidVerifyingKey, InvalidSigningKey
@@ -122,14 +122,12 @@ class TransactionsEndpoint(NamespacedDriver):
         verifying_key = verifying_key if verifying_key else self.verifying_key
         if not verifying_key:
             raise InvalidVerifyingKey
-        fulfillment = Fulfillment.gen_default([verifying_key])
-        condition = fulfillment.gen_condition()
-        data = Data(payload=payload)
-        transaction = Transaction(
-            'CREATE',
-            fulfillments=[fulfillment],
-            conditions=[condition],
-            data=data,
+        # TODO: In the future, we should adjust this to let the user of the
+        #       driver define both `owners_before` and `owners_after`.
+        transaction = Transaction.create(
+            owners_before=[verifying_key],
+            owners_after=[verifying_key],
+            payload=payload
         )
         signed_transaction = transaction.sign([signing_key])
         return self._push(signed_transaction.to_dict())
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -81,26 +80,9 @@ def mockreturn(*args, **kwargs):
...
-    return Transaction(Transaction.CREATE,
-                       fulfillments=[fulfillment],
-                       conditions=[condition])
-
-
-@fixture
-def bob_condition(bob_pubkey):
-    condition_uri = Ed25519Fulfillment(public_key=bob_pubkey).condition_uri
-    return Condition(condition_uri, owners_after=[bob_pubkey])
+def transaction(alice_pubkey):
```
sohkai: `alice_transaction` is probably a better name.

bigchaindb_driver/driver.py
```diff
@@ -183,7 +181,11 @@ def transfer(self, transaction, *conditions, signing_key=None):
         signing_key = signing_key if signing_key else self.signing_key
         if not signing_key:
             raise InvalidSigningKey
-        transfer_transaction = transaction.transfer(list(conditions))
+        inputs = transaction.to_inputs()
```
sohkai: I think `get_inputs()` is a better name than `to_inputs()` (I assume this just extracts the inputs out of the transaction?).

bigchaindb_driver/driver.py
```diff
@@ -122,14 +122,12 @@ def create(self, payload=None, verifying_key=None, signing_key=None):
...
-        data = Data(payload=payload)
-        transaction = Transaction(
-            'CREATE',
-            fulfillments=[fulfillment],
-            conditions=[condition],
-            data=data,
+        # TODO: In the future, we should adjust this to let the user of the
+        #       driver define both `owners_before` and `owners_after`.
+        transaction = Transaction.create(
+            owners_before=[verifying_key],
```
sohkai: Does `owners_before` make sense when creating transactions? Is it necessary (shouldn't the node that creates the transaction fill that in)?

tests/conftest.py
TimDaub: OK but this is out of the scope of this PR.
I'll add a TODO.

bigchaindb_driver/driver.py
TimDaub: `get_inputs` would be arbitrary since `transaction` could be of `operation == TRANSFER`, so `get_inputs` would mean you retrieve the already spent inputs :)

bigchaindb_driver/driver.py
TimDaub: > Does owners_before make sense when creating transactions?

No not really. Anything would go through I guess. 

> Is it necessary (shouldn't the node that creates the transaction fill that in)?

This is how it's being done at this point, but we actually wanted to get rid of that.
I guess only `owners_after` would be necessary at one point.

bigchaindb_driver/driver.py
sohkai: Mmm, I see. This naming can get really confusing...

sohkai:
Looks good!

PR #64, opened 2016-08-29
Rename BigchaindbException to DriverException

sohkai:


sbellem:
Perhaps, `DriverException` is too generic. From a developer's standpoint who may be working with other backends, (e.g.: [redis](https://github.com/andymccurdy/redis-py/blob/master/redis/exceptions.py#L5), [elasticsearch](https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/exceptions.py#L13), [postgres](https://github.com/MagicStack/asyncpg/blob/master/asyncpg/exceptions/_base.py#L96)), seeing `DriverException` will not indicate right away from which client/driver the error is originating.

sohkai:
Good points, and certainly with how Python shows the exception, it's unclear where it came from. I'll revert this change in a PR.

PR #72, opened 2016-09-09
Add sanity checks to given keypairs during driver initialization

sohkai:
Some small sanity checks for making sure the keys are strings and that if a signing key is given, a verifying key is also given.

Not sure if it'll ever be useful to only give a signing key, but I imagine the other way around (only binding a verifying key to the driver) may be useful for people to only supply their signing key when absolutely necessary.

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -37,8 +37,8 @@ class BigchainDB:
                  transport_class=Transport):
         """Initialize a :class:`~bigchaindb_driver.BigchainDB` driver instance.
 
-        If a :attr:`verifying_key` and :attr:`signing_key` are given, this
-        instance will be bound to the given keypair and be applied as defaults
+        If a :attr:`verifying_key` or :attr:`signing_key` are given, this
+        instance will be bound to the keys and applied them as defaults
         whenever a verifying and/or signing key are needed.
 
         Args:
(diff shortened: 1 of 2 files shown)


sbellem:
I wonder if this is not too much validation here, and if we could instead fall back on the errors being thrown by bigchaindb-common when keys are not correct.

bigchaindb_driver/driver.py
```diff
@@ -37,19 +37,41 @@ def __init__(self,
...
         """
+        if verifying_key is not None and not isinstance(verifying_key, str):
+                raise InvalidVerifyingKey("'verifying_key' must be a string")
+
+        if signing_key is not None:
+            if not isinstance(signing_key, str):
+                raise InvalidSigningKey("'signing_key must be a string")
+            if verifying_key is None:
+                raise InvalidVerifyingKey(("'verifying_key' must be given if"
+                                           "'signing_key' is given"))
```
sbellem: Using `.transfer` does not require a `verifying_key`, so why not allow a user to only supply a `signing_key`?

tests/test_driver.py
```diff
@@ -16,11 +16,41 @@ def test_temp_driver_returns_a_temp_driver(bdb_node):
...
+    with raises(InvalidSigningKey):
+        BigchainDB(verifying_key=alice_keypair.vk, signing_key=0)
+
+    # A verifying key is necessary when binding a signing_key
+    with raises(InvalidVerifyingKey):
+        BigchainDB(signing_key=alice_keypair.sk)
+    # But not the other way around
+    driver = BigchainDB(verifying_key=alice_keypair.vk)
+    assert driver.verifying_key == alice_keypair.vk
+    assert driver.signing_key is None
```
sbellem: This test could be divided into 4 tests.

bigchaindb_driver/driver.py
sohkai: The `verifying_key` is public, so I'm assuming people are ok with binding the driver to that. They might not want to expose their private key like this though (and might also only want to load it from disk / etc when they need it), and so I left it out. In fact, there's no use case in the driver right now that only depends on the `verifying_key`.

PR #74, opened 2016-09-12
Expand connection layer interfaces

sohkai:
Expands ability to customize the transport layers, with predefined abstract interfaces to implement.

Adds ability to specify these customizations at the driver level, by using kwargs.

### DIFF
--- a/bigchaindb_driver/connection.py
+++ b/bigchaindb_driver/connection.py
@@ -10,7 +10,13 @@ HttpResponse = namedtuple('HttpResponse', ('status_code', 'headers', 'data'))
 
 
 class AbstractConnection(ABC):
-    """Abstract interface for Connection classes"""
+    """Abstract interface for Connection classes
+
+    Attributes:
+        node_url (str): URL of this connection
+        session (:class:`~requests.Session`): Long-lived connection to
+            :attr:`node_url`
+    """
 
     @abstractmethod
     def __init__(self, node_url):
--- a/bigchaindb_driver/connection.py
+++ b/bigchaindb_driver/connection.py
@@ -1,3 +1,4 @@
+from abc import ABC, abstractmethod
 from collections import namedtuple
 
 from requests import Session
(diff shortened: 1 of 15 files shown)


sbellem:
@sohkai could you please put typo fixes and documentation enhancements into separate PRs?

bigchaindb_driver/connection.py
```diff
@@ -8,29 +9,72 @@
...
+            :attr:`node_url`
+    """
+
+    @abstractmethod
+    def __init__(self, node_url):
+        """Initialize a Connnection class to the given node
 
         Args:
-            node_url (str):  Url of the node to connect to.
+            node_url (str): URL of the node to connect to
```
sohkai: Other areas used `URL`, we can pick one or the other and stick to it.

bigchaindb_driver/connection.py
```diff
@@ -8,29 +9,72 @@
...
+
+        See :meth:`~bigchaindb_driver.connection.AbstractConnection.__init__`
+        for the arguments.
         """
         self.node_url = node_url
         self.session = Session()
 
-    def request(self, method, *, path=None, json=None, **kwargs):
-        """Performs an HTTP requests for the specified arguments.
+    def request(self, method, path=None, *, json=None, **kwargs):
```
sohkai: See commit message in https://github.com/bigchaindb/bigchaindb-driver/pull/74/commits/7dc6958557db3d416c9a6470c9d9f08f6aa3b044

bigchaindb_driver/driver.py
```diff
@@ -1,57 +1,75 @@
...
 
         If a :attr:`verifying_key` or :attr:`signing_key` are given, this
         instance will be bound to the keys and applied them as defaults
         whenever a verifying and/or signing key are needed.
 
         Args:
-            *nodes (str): BigchainDB nodes to connect to. Currently, the full
-                URL must be given. In the absence of any node, the default of
-                the :attr:`transport_class` will be used, e.g.:
+            *node_urls (str): BigchainDB nodes to connect to. Currently, the
```
sohkai: Right now every node connection is a url, but if this will change later, we can revert this.

bigchaindb_driver/driver.py
```diff
@@ -1,57 +1,75 @@
...
+                full URL must be given. In the absence of any node, the default
+                of the :attr:`transport_cls` will be used, e.g.:
                 ``'http://localhost:9984/api/v1'``.
             verifying_key (str, keyword, optional): the base58 encoded public
                 key for the ED25519 curve to bind this driver with.
             signing_key (str, keyword, optional): the base58 encoded private
                 key for the ED25519 curve to bind this driver with.
-            transport_class (Transport, keyword, optional): Transport class to
-                use.
+            transport_cls (Transport, keyword, optional): a subclass of
```
sohkai: Purely stylistic since `cls` is used for classmethods, but if it's more common to use `class` when passing in classes, we can do so.

sbellem:
@sohkai regarding e3d728d1efbd462559f440fe164926b9db0f2f22, the renaming of `nodes` to `node_url`: the name `nodes` was chosen with the plan of supporting multiple ways to input nodes (e.g.: `hostname`, `hostname:port`, `scheme://hostname:port`). For this reason `nodes` is more appropriate, whereas `node_urls` only addresses one of the eventually supported cases.

sbellem:
@sohkai more generally speaking, please split the PR in such a way that different concerns are addressed by different PRs.

sohkai:
Will break out into smaller PRs.

PR #89, opened 2016-09-29
Add docs for usage

sbellem:


docs/usage.rst
```diff
@@ -83,13 +95,13 @@ and then transfers it to Bob:
```
sohkai: Should be `creation_tx = bdb.transactions.retrieve(creation_tx['id'])`

docs/usage.rst
```diff
@@ -113,12 +125,49 @@ Bob is the new owner:
...
+
+.. code-block:: python
+
+    import logging
+
+    from bigchaindb_driver import BigchainDB
+    from bigchaindb_driver.exceptions import NotFoundError
+
+    logger = logging.getLogger(__name__)
+    logging.basicConfig(format='%(asctime)-15s %(status)-3s %(message)s')
```
sohkai: Maybe just use `print` here to make it less complicated? If they want to use the logger, I'm sure they'll be able to figure out how elsewhere.

sbellem: :smile: This was really intentional!

sohkai: I could tell ;).

sohkai:
LGTM; we can move the error handling to its own section later.

PR #92, opened 2016-10-05
Closes #86

### LINKED ISSUE #86: Support payload parameter for transfer

sbellem:


sohkai:
Noticed this failed likely because BDB took longer than 1.5s to validate the transfer transaction. I use a [small polling utility](https://github.com/bigchaindb/pycoalaip-bigchaindb/blob/master/tests/utils.py#L6) to check up to 5 seconds in my tests that might be more robust than a single sleep. Looks good otherwise.

But man, does pytest-sugar look nice on travis.

sbellem:
@sohkai it is a double spend:

```bash
bdb-server_1  | bigchaindb_common.exceptions.DoubleSpend: `64908c64de78cc617ccefaac7f0657d550a90e98826bb3441899c33757feed31` was spent more then once. There is a problem with the chain
```

tests/conftest.py
```diff
@@ -86,8 +87,12 @@ def mockreturn(*args, **kwargs):
 
 @fixture
 def alice_transaction_obj(alice_pubkey):
-    return Transaction.create(owners_before=[alice_pubkey],
-                              owners_after=[alice_pubkey])
+    serial_number = b64encode(urandom(10), altchars=b'-_').decode()
```
sohkai: Creating a random value here shouldn't be necessary for a payload uuid; but if you like using it then 👍.

sohkai:
👍

PR #116, opened 2016-10-18
Fix and update driver to match latest server state

sbellem:


sohkai:
LGTM for now as a step towards compatibility with the current version of BDB.

PR #121, opened 2016-10-26
120

sbellem:
addresses #120


setup.py
```diff
@@ -39,6 +38,10 @@
     'sphinxcontrib-httpdomain',
 ]
 
+dependency_links = [
```
sohkai: Let's just wait until 0.7 is out.

sohkai:
LGTM.

PR #135, opened 2016-11-09
Removed HISTORY.rst file and changed dev status to Alpha

ttmc:
Trent said we should change the status of the BigchainDB Python Driver from pre-alpha to alpha, so I did that.

I also deleted the `HISTORY.rst` file and everything that referenced it. It's a nice idea, but we have better things to do than maintain that (something we weren't doing anyway). Git has the code history and if you really want a list of all `bigchaindb_driver` packages that PyPI has ever had, see:

https://pypi.python.org/simple/bigchaindb-driver/


TimDaub:
I can recommend using [travis pypi integration with Github releases](https://github.com/bigchaindb/bigchaindb-driver/blob/master/travis_pypi_setup.py) as [I did it with bigchaindb_common](https://github.com/bigchaindb/bigchaindb-common/releases).

HISTORY.rst
```diff
@@ -1,13 +0,0 @@
...
-
-0.0.1a1 (2016-08-12)
---------------------
-
-* Planning release on PyPI.
-
-0.0.1dev1 (2016-08-25)
-----------------------
-
-* Development (pre-alpha) release on PyPI.
```
sbellem: The purpose of `HISTORY.rst` was to log important changes. It has not been used so far as it was judged unnecessary meanwhile the driver is under the kind of development it was under. We will need it soon probably, especially if we introduce breaking changes, or significant changes. 

So what is the plan now?

sbellem: Moreover, it came at no cost as it was generated by the cookiecutter tool.

sbellem:
@ttmc The one commit addresses two very separate and different topics. I understand that they both may be very minor, but in general, it is good practice to have commits that address one thing.

TimDaub:
> Is this related to this PR in anyway?

Yes. You can log (breaking) changes in the notes of a release.

sbellem:
> Yes. You can log (breaking) changes in the notes of a release.

Ok. But the first question that will need to be addressed before doing so is whether we do want to do automated releases for the driver. As of now, the answer to this question was no. We can revisit the question, but  this is a separate issue. 

And for now there is a working boilerplate, which has been put into place, and I do not see reasons for changing it at this very moment. The focus should be on the development of the driver itself, not on its boilerplate, which is really solid, and based on best practices from the community.

ttmc:
I will close this PR and I'll make a new one where I only change the status to Alpha.

PR #138, opened 2016-11-09
Separate docs for quick and advanced installation

ttmc:
* Created a new page: **Quickstart Installation**. It has some quick installation recipes for Ubuntu and Fedora.
* Change the old **Installation** page to **Advanced Installation** and expanded it a bit.
* Re-titled the **Usage** page to **Basic Usage Examples**. The file is still named `usage.rst`. Also added some notes to the top of that page.
* Updated the Sphinx index.


vrde:
:+1: for the separation. Some devs are just looking for **copypasta**.

sbellem:
I think we should agree on who our target audience is.

From the point of view of many developers, this separation may seem to be overly complicated, and potentially confusing, at this point in time. A quickstart installation, if we have one, could simply be:

### From PyPI
```bash
$ pip install bigchaindb_driver
```

### From source
... as it is now --> http://bigchaindb.readthedocs.io/projects/py-driver/en/latest/installation.html#installing-from-the-source-code

Anyhow, the bottom line is that we need to agree on what is the assumed knowledge of our audience,

My opinion on this is that we are assuming too less in this PR.

As an example, if we look at [requests](http://docs.python-requests.org/en/master/user/install/#install) it is straight to the point. Why not the same with BigchainDB, is a question worth asking I think.

Another example to look at is RethinkDB: https://www.rethinkdb.com/docs/install-drivers/python/

Moreover, the dependencies relating to cryptoconditions could, and should actually be documented in cryptoconditions, and the driver may point to these dependencies, or if we wish to repeat them, then ok.

The instructions on how to install python, pip, and so forth if we include them, could be under a section/page titled something like "New to Python".

sbellem:
Note: I have added (as assignees) @rhsimplex @r-marques and @vrde to help in in clarifying the question:

"Who is our target audience, and is the knowledge base we assume they have?"

I am aware that this question goes beyond this PR, and that this PR is more or less in line with other BigchainDB docs. 

Consequently, if you all feel that we could move the question to a separate "general" issue, it's fine with me.

ttmc:
There's an **Advanced Installation** page for people who want _more_. The **Quickstart Installation** is designed to be quick and mindless (hence the word _Quickstart_). It should just work, provided the reader can find their way to a command line. In my experience, it's much easier to lose someone than it is to insult their intelligence.

@sbellem Maybe create a separate issue asking who our audiences are. We can post our thoughts about that there. The docs can always be improved. I'd just like to get this PR merged.

vrde:
> My opinion on this is that we are assuming too less in this PR.

Well, OK maybe we can remove the part "To start an interactive Python 3 session, use: python3".
On the other side I think it's good to be *pedantic* and make sure the user is on Python 3 and has `setuptools` and `pip` up-to-date.

> Moreover, the dependencies relating to cryptoconditions could, and should actually be documented in cryptoconditions, and the driver may point to these dependencies, or if we wish to repeat them, then ok.

Our driver depends on cryptoconditions, hence the installation requires also a compiler, IMO Troy is saving a lot of time to developers here :hourglass:. Check out also the instructions to install [pymongo](http://api.mongodb.com/python/current/installation.html). They include also dependencies to install.

A navigated Python user can ignore the extra info, while a beginner might save hours.

sbellem:
> There's an Advanced Installation page for people who want more. The Quickstart Installation is designed to be quick and mindless (hence the word Quickstart). It should just work, provided the reader can find their way to a command line. In my experience, it's much easier to lose someone than it is to insult their intelligence.

To re-emphasize the point I was trying to make: there is an audience for whom neither the quickstart page nor the advanced page will suit. That is why I was pointing at requests, and RethinkDB. [MongoDB](https://api.mongodb.com/python/current/installation.html) as well, gives it to you right away. That essential part is:

```bash
$ pip install bigchaindb_driver
```

Moreover we are totally neglecting the Mac OS X users with the approach in this PR

Moreover, a new user when landing on the docs, is presented the choice:

* Quickstart 

OR

* Advanced Usage

It's a gamble.

Mac OS X users for instance, if they choose Quickstart: oh bad luck! Need to go to Advanced Usage, and then need to go through OS specific docs, to finally find http://bigchaindb.readthedocs.io/projects/py-driver/en/latest/advanced-installation.html#installing-the-bigchaindb-python-driver

I think this is overly complicated for what is needed in the end. Why not follow in the steps of other successful Python projects?

PR #143, opened 2016-11-17
Provide separate functions for creating tx payloads, and fulfilling them

sbellem:
**IMPORTANT**: targets BigchainDB `0.7.x`

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -2,6 +2,7 @@ from bigchaindb.common.transaction import Asset, Transaction
 
 from .exceptions import InvalidVerifyingKey, InvalidSigningKey
 from .transport import Transport
+from .offchain import prepare_transaction, fulfill_transaction
 
 
 DEFAULT_NODE = 'http://localhost:9984/api/v1'
@@ -119,6 +120,131 @@ class TransactionsEndpoint(NamespacedDriver):
     """
     path = '/transactions/'
 
+    @staticmethod
+    def prepare(*, operation='CREATE', owners_before=None,
+                owners_after=None, asset=None, metadata=None, inputs=None):
+        """
+        Prepares a transaction payload, ready to be fulfilled.
+
+        Args:
+            operation (str): The operation to perform. Must be ``'CREATE'``
+                or ``'TRANSFER'``. Case insensitive. Defaults to ``'CREATE'``.
+            owners_before (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+                One or more public keys representing the issuer(s) of
+                the asset being created. Only applies for ``'CREATE'``
+                operations. Defaults to ``None``.
+            owners_after (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+                One or more public keys representing the new owner(s) of the
+                asset being created or transferred. Defaults to ``None``.
+            asset (:obj:`dict`, optional): The asset being created or
+                transferred. MUST be supplied for ``'TRANSFER'`` operations.
+                Defaults to ``None``.
+            metadata (:obj:`dict`, optional): Metadata associated with the
+                transaction. Defaults to ``None``.
+            inputs (:obj:`dict` | :obj:`list` | :obj:`tuple`, optional):
+                One or more inputs holding the condition(s) that this
+                transaction intends to fulfill. Each input is expected to
+                be a :obj:`dict`. Only applies to, and MUST be supplied for,
+                ``'TRANSFER'`` operations.
+
+        Returns:
+            dict: The prepared transaction.
+
+        Raises:
+            :class:`~.exceptions.BigchaindbException`: If ``operation`` is
+                not ``'CREATE'`` or ``'TRANSFER'``.
+
+        .. important::
+
+            **CREATE operations**
+
+            * ``owners_before`` MUST be set.
+            * ``owners_after``, ``asset``, and ``metadata`` MAY be set.
+            * The argument ``inputs`` is ignored.
+            * If ``owners_after`` is not given, or evaluates to
+              ``False``, it will be set equal to ``owners_before``::
+
+                if not owners_after:
+                    owners_after = owners_before
+
+            **TRANSFER operations**
+
+            * ``owners_after``, ``asset``, and ``inputs`` MUST be set.
+            * ``metadata`` MAY be set.
+            * The argument ``owners_before`` is ignored.
+
+        """
+        return prepare_transaction(
+            operation=operation,
+            owners_before=owners_before,
+            owners_after=owners_after,
+            asset=asset,
+            metadata=metadata,
+            inputs=inputs,
+        )
+
+    @staticmethod
+    def fulfill(transaction, private_keys):
+        """
+        Fulfills the given transaction.
+
+        Args:
+            transaction (dict): The transaction to be fulfilled.
+            private_keys (:obj:`str` | :obj:`list` | :obj:`tuple`): One or
+                more private keys to be used for fulfilling the
+                transaction.
+
+        Returns:
+            dict: The fulfilled transaction payload, ready to be sent to a
+            BigchainDB federation.
+
+        Raises:
+            :exception:`~.exceptions.MissingSigningKeyError`: If a private
+                key, (aka signing key), is missing.
+
+        """
+        return fulfill_transaction(transaction, private_keys=private_keys)
+
+    def send(self, transaction):
+        """Submit a transaction to the Federation.
+
+        Args:
+            transaction (dict): the transaction to be sent
+                to the Federation node(s).
+
+        Returns:
+            dict: The transaction sent to the Federation node(s).
+
+        """
+        return self.transport.forward_request(
+            method='POST', path=self.path, json=transaction)
+
+    def retrieve(self, txid):
+        """Retrieves the transaction with the given id.
+
+        Args:
+            txid (str): Id of the transaction to retrieve.
+
+        Returns:
+            dict: The transaction with the given id.
+
+        """
+        path = self.path + txid
+        return self.transport.forward_request(method='GET', path=path)
+
+    def status(self, txid):
+        """Retrieves the status of the transaction with the given id.
+
+        Args:
+            txid (str): Id of the transaction to retrieve the status for.
+
+        Returns:
+            dict: A dict containing a 'status' item for the transaction.
+
+        """
+        path = self.path + txid + '/status'
+        return self.transport.forward_request(method='GET', path=path)
+
     def create(self, asset=None, verifying_key=None, signing_key=None):
         """Issue a transaction to create an asset.
 
@@ -155,33 +281,7 @@ class TransactionsEndpoint(NamespacedDriver):
             asset=asset,
         )
         signed_transaction = transaction.sign([signing_key])
-        return self._push(signed_transaction.to_dict())
-
-    def retrieve(self, txid):
-        """Retrieves the transaction with the given id.
-
-        Args:
-            txid (str): Id of the transaction to retrieve.
-
-        Returns:
-            dict: The transaction with the given id.
-
-        """
-        path = self.path + txid
-        return self.transport.forward_request(method='GET', path=path)
-
-    def status(self, txid):
-        """Retrieves the status of the transaction with the given id.
-
-        Args:
-            txid (str): Id of the transaction to retrieve the status for.
-
-        Returns:
-            dict: A dict containing a 'status' item for the transaction.
-
-        """
-        path = self.path + txid + '/status'
-        return self.transport.forward_request(method='GET', path=path)
+        return self.send(signed_transaction.to_dict())
 
     def transfer(self, transaction, *owners_after, asset, signing_key=None):
         """Issue a transaction to transfer an asset.
@@ -209,17 +309,4 @@ class TransactionsEndpoint(NamespacedDriver):
             list(owners_after),
             asset=Asset.from_dict(asset),
         ).sign([signing_key]).to_dict()
-        return self._push(signed_transfer_transaction)
-
-    def _push(self, transaction):
-        """Submit a transaction to the Federation.
-
-        Args:
-            transaction (dict): the transaction to be pushed to the Federation.
-
-        Returns:
-            dict: The transaction pushed to the Federation.
-
-        """
-        return self.transport.forward_request(
-            method='POST', path=self.path, json=transaction)
+        return self.send(signed_transfer_transaction)
--- a/bigchaindb_driver/offchain.py
+++ b/bigchaindb_driver/offchain.py
@@ -1,0 +1,328 @@
+"""
+Module for operations that can be performed "offchain", meaning without
+a connection to one or more  BigchainDB federation nodes.
+
+"""
+import logging
+from functools import singledispatch
+
+from bigchaindb.common.transaction import (
+    Fulfillment as BdbFulfillment,
+    Transaction,
+    TransactionLink,
+)
+from bigchaindb.common.exceptions import KeypairMismatchException
+from cryptoconditions import Fulfillment
+
+from .exceptions import BigchaindbException, MissingSigningKeyError
+from .utils import (
+    CreateOperation,
+    TransferOperation,
+    _normalize_asset,
+    _normalize_operation,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@singledispatch
+def _prepare_transaction(operation,
+                         owners_before=None,
+                         owners_after=None,
+                         asset=None,
+                         metadata=None,
+                         inputs=None):
+    raise BigchaindbException((
+        'Unsupported operation: {}. '
+        'Only "CREATE" and "TRANSFER" are supported.'.format(operation)))
+
+
+@_prepare_transaction.register(CreateOperation)
+def _prepare_create_transaction_dispatcher(operation, **kwargs):
+    del kwargs['inputs']
+    return prepare_create_transaction(**kwargs)
+
+
+@_prepare_transaction.register(TransferOperation)
+def _prepare_transfer_transaction_dispatcher(operation, **kwargs):
+    del kwargs['owners_before']
+    return prepare_transfer_transaction(**kwargs)
+
+
+def prepare_transaction(*, operation='CREATE', owners_before=None,
+                        owners_after=None, asset=None, metadata=None,
+                        inputs=None):
+    """
+    Prepares a transaction payload, ready to be fulfilled. Depending on
+    the value of ``operation`` simply dispatches to either
+    :func:`~.prepare_create_transaction` or
+    :func:`~.prepare_transfer_transaction`.
+
+    Args:
+        operation (str): The operation to perform. Must be ``'CREATE'``
+            or ``'TRANSFER'``. Case insensitive. Defaults to ``'CREATE'``.
+        owners_before (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+            One or more public keys representing the issuer(s) of the
+            asset being created. Only applies for ``'CREATE'``
+            operations. Defaults to ``None``.
+        owners_after (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+            One or more public keys representing the new owner(s) of the
+            asset being created or transferred. Defaults to ``None``.
+        asset (:obj:`dict`, optional): The asset being created or
+            transferred. MUST be supplied for ``'TRANSFER'`` operations.
+            Defaults to ``None``.
+        metadata (:obj:`dict`, optional): Metadata associated with the
+            transaction. Defaults to ``None``.
+        inputs (:obj:`dict` | :obj:`list` | :obj:`tuple`, optional):
+            One or more inputs holding the condition(s) that this
+            transaction intends to fulfill. Each input is expected to
+            be a :obj:`dict`. Only applies to, and MUST be supplied for,
+            ``'TRANSFER'`` operations.
+
+    Returns:
+        dict: The prepared transaction.
+
+    Raises:
+        :class:`~.exceptions.BigchaindbException`: If ``operation`` is
+            not ``'CREATE'`` or ``'TRANSFER'``.
+
+    .. important::
+
+        **CREATE operations**
+
+        * ``owners_before`` MUST be set.
+        * ``owners_after``, ``asset``, and ``metadata`` MAY be set.
+        * The argument ``inputs`` is ignored.
+        * If ``owners_after`` is not given, or evaluates to
+          ``False``, it will be set equal to ``owners_before``::
+
+            if not owners_after:
+                owners_after = owners_before
+
+        **TRANSFER operations**
+
+        * ``owners_after``, ``asset``, and ``inputs`` MUST be set.
+        * ``metadata`` MAY be set.
+        * The argument ``owners_before`` is ignored.
+
+    """
+    operation = _normalize_operation(operation)
+    return _prepare_transaction(
+        operation,
+        owners_before=owners_before,
+        owners_after=owners_after,
+        asset=asset,
+        metadata=metadata,
+        inputs=inputs,
+    )
+
+
+def prepare_create_transaction(*,
+                               owners_before,
+                               owners_after=None,
+                               asset=None,
+                               metadata=None):
+    """
+    Prepares a ``"CREATE"`` transaction payload, ready to be
+    fulfilled.
+
+    Args:
+        owners_before (:obj:`list` | :obj:`tuple` | :obj:`str`): One
+            or more public keys representing the issuer(s) of the asset
+            being created.
+        owners_after (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+            One or more public keys representing the new owner(s) of the
+            asset being created. Defaults to ``None``.
+        asset (:obj:`dict`, optional): The asset being created. Defaults
+            to ``None``.
+        metadata (:obj:`dict`, optional): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"CREATE"`` transaction.
+
+    .. important:: If ``owners_after`` is not given, or evaluates to
+        ``False``, it will be set equal to ``owners_before``::
+
+            if not owners_after:
+                owners_after = owners_before
+
+    """
+    if not isinstance(owners_before, (list, tuple)):
+        owners_before = [owners_before]
+
+    # NOTE: Needed for the time being. See
+    # https://github.com/bigchaindb/bigchaindb/issues/797
+    if isinstance(owners_before, tuple):
+        owners_before = list(owners_before)
+
+    if not owners_after:
+        owners_after = owners_before
+
+    asset = _normalize_asset(asset)
+
+    transaction = Transaction.create(
+        owners_before,
+        owners_after,
+        metadata=metadata,
+        asset=asset,
+    )
+    return transaction.to_dict()
+
+
+def prepare_transfer_transaction(*,
+                                 inputs,
+                                 owners_after,
+                                 asset,
+                                 metadata=None):
+    """
+    Prepares a ``"TRANSFER"`` transaction payload, ready to be
+    fulfilled.
+
+    Args:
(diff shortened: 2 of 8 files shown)


bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+    CreateOperation,
+    TransferOperation,
+    _normalize_asset,
+    _normalize_operation,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@singledispatch
```
diminator: not sure if this adds to readiblity...

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+    if not isinstance(owners_before, (list, tuple)):
+        owners_before = [owners_before]
+
+    # NOTE: Needed for the time being. See
+    # https://github.com/bigchaindb/bigchaindb/issues/797
+    if isinstance(owners_before, tuple):
+        owners_before = list(owners_before)
+
+    if not owners_after:
+        owners_after = owners_before
```
diminator: explain?

bigchaindb_driver/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    try:
+        operation = ops_map[operation]()
+    except KeyError:
+        pass
+
+    return operation
+
+
+def _normalize_asset(asset, is_transfer=False):
+    if is_transfer:
```
diminator: this routine is a bit unclear

From the input arg `asset` you want to make sure you are returning an `Asset` object
One can pass either a `dict` or a list of `kwargs`?
The switch is based on `is_transfer`? Not on type of the arg `asset`?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        dict: The prepared ``"TRANSFER"`` transaction.
+
+    """
+    if not isinstance(inputs, (list, tuple)):
+        inputs = (inputs, )
+    if not isinstance(owners_after, (list, tuple)):
+        owners_after = [owners_after]
+
+    fulfillments = [
+        BdbFulfillment(Fulfillment.from_dict(input_['fulfillment']),
```
diminator: `Fulfillment.from_dict(input_['fulfillment'])`: why referring to the `fulfillment` of the `input` here?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # NOTE: Needed for the time being. See
+    # https://github.com/bigchaindb/bigchaindb/issues/797
+    if isinstance(owners_before, tuple):
+        owners_before = list(owners_before)
+
+    if not owners_after:
+        owners_after = owners_before
+
+    asset = _normalize_asset(asset)
```
diminator: why is there normalization needed?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        owners_after = [owners_after]
+
+    fulfillments = [
+        BdbFulfillment(Fulfillment.from_dict(input_['fulfillment']),
+                       input_['owners_before'],
+                       tx_input=TransactionLink(**input_['input']))
+        for input_ in inputs
+    ]
+
+    asset = _normalize_asset(asset, is_transfer=True)
```
diminator: why is this normalization different?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+                                 metadata=None):
+    """
+    Prepares a ``"TRANSFER"`` transaction payload, ready to be
+    fulfilled.
+
+    Args:
+        inputs (:obj:`dict` | :obj:`list` | :obj:`tuple`): One or more
+            inputs holding the condition(s) that this transaction
+            intents to fulfill. Each input is expected to be a
+            :obj:`dict`.
```
diminator: structure of the `dict`? `{txid: <ID>, cid: <int>}`?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+                               owners_before,
+                               owners_after=None,
+                               asset=None,
+                               metadata=None):
+    """
+    Prepares a ``"CREATE"`` transaction payload, ready to be
+    fulfilled.
+
+    Args:
+        owners_before (:obj:`list` | :obj:`tuple` | :obj:`str`): One
```
diminator: why are there `owners_before`? Isn't this always the/a federation/node?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        owners_after,
+        asset,
+        metadata=metadata,
+    )
+    return transaction.to_dict()
+
+
+def fulfill_transaction(transaction, *, private_keys):
+    """
+    Fulfills the given transaction.
```
diminator: only for multisig?

bigchaindb_driver/offchain.py
sbellem: Oh, this was changed. I think the change was made under https://github.com/bigchaindb/bigchaindb/commit/50b0b3cef217bc412dba5a062de1db2d1adcd514 but not sure.

@TimDaub and @r-marques may have more insights on this

bigchaindb_driver/offchain.py
sbellem: This is to free the user from passing redundant information. 

The way the `CREATE` operation works is that the `owners_before` refers to the original owner(s).  They can make an implicit transfer by setting `owners_after` to different public keys.

The common api requires to pass both `owners_before` and `owners_after` even if they are the same.

bigchaindb_driver/offchain.py
sbellem: The common api requires an `Asset` instance, whereas the driver apis work strictly with `dict` objects.

So, internally, the `dict` needs to be converted to an `Asset`instance.

This also helps to isolate dependencies on `common`.

bigchaindb_driver/offchain.py
r-marques: This changed. Now any client can create assets.
Before we enforced that only federation nodes could create assets but we never had any custom rules for this. We would just accept any create request from clients.
This also takes care of improving the security of create transactions, since the client signs the create transaction no one else can change its content

bigchaindb_driver/offchain.py
sbellem: I assume you are suggesting it could be documented? That would be a good point.

It is the complete outer structure. Say you have a `tx` payload you are using as an "input":

```python
cid = 0
condition = tx['transaction']['conditions'][cid]
{
    'fulfillment': condition['condition']['details'],
    'input': {
        'cid': cid,
        'txid': tx['id'],
    },
    'owners_before': condition['owners_after'],
}
```

bigchaindb_driver/driver.py
```diff
@@ -119,43 +120,91 @@ class TransactionsEndpoint(NamespacedDriver):
...
-            :class:`~bigchaindb_driver.exceptions.InvalidVerifyingKey`: If
-                neither ``verifying_key`` nor ``self.verifying_key`` have
-                been set.
+            :class:`~.exceptions.BigchaindbException`: If ``operation`` is
+                not ``'CREATE'`` or ``'TRANSFER'``.
+
+        .. important::
+
+            * For ``operation='CREATE'``, ``owners_before`` MUST be set,
+              meanwhile ``owners_after``, ``asset``, amd ``metadata`` MAY be
```
r-marques: `amd` -> `and`

bigchaindb_driver/offchain.py
sbellem: This is due to how the `common` interface works. For `CREATE` operations it is not necessary to supply an asset `id` whereas for `TRANSFER` operations, an asset `id` must be passed. Moreover the way to instantiate an `Asset` object can be slightly different depending on the operation type, hence, the parameter `is_transfer`.

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        dict: The prepared transaction.
+
+    Raises:
+        :class:`~.exceptions.BigchaindbException`: If ``operation`` is
+            not ``'CREATE'`` or ``'TRANSFER'``.
+
+    .. important::
+
+        * For ``operation='CREATE'``, ``owners_before`` MUST be set,
+          meanwhile ``owners_after``, ``asset``, amd ``metadata`` MAY be
```
r-marques: `amd` -> `and`

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+          set. The argument ``inputs`` is ignored.
+
+        * For ``operation='TRANSFER'``, ``owners_after``, ``asset``,
+          and ``inputs`` MUST be set, meanwhile  ``mwtadata`` MAY be set.
```
r-marques: `mwtadata` -> `metadata`

bigchaindb_driver/utils.py
sbellem: Not sure I fully understand your questions. Here are a few points that may help to clarify:

1. The function is purely for internal use.
2. The function is to help isolate dependencies on `common`.
3. The function is meant to take a `dict` representing an asset, and convert it to a normalized form, which for the time being is an `Asset` instance.

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+                               owners_after=None,
+                               asset=None,
+                               metadata=None):
+    """
+    Prepares a ``"CREATE"`` transaction payload, ready to be
+    fulfilled.
+
+    Args:
+        owners_before (:obj:`list` | :obj:`tuple` | :obj:`str`): One
+            or more public keys representing the original owner(s) of
```
r-marques: Instead of `the original owner(s) of` we could use `the issuer of`

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+            asset being created. Defaults to ``None``.
+        asset (:obj:`dict`, optional): The asset being created. Defaults
+            to ``None``.
+        metadata (:obj:`dict`, optional): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"CREATE"`` transaction.
+
+    """
```
r-marques: Its worth adding and `.. important::` here saying that if `owners_after` is not specified, it will be set to the same as `owners_before`

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        metadata (:obj:`dict`): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"TRANSFER"`` transaction.
+
+    """
+    if not isinstance(inputs, (list, tuple)):
+        inputs = (inputs, )
+    if not isinstance(owners_after, (list, tuple)):
```
r-marques: won't this have the same problem as https://github.com/bigchaindb/bigchaindb/issues/797?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+        owners_after = [owners_after]
+
+    fulfillments = [
+        BdbFulfillment(Fulfillment.from_dict(input_['fulfillment']),
+                       input_['owners_before'],
```
r-marques: Shouldn't this be the `owners_after` of the condition?

bigchaindb_driver/offchain.py
sbellem: Ah! Since https://github.com/bigchaindb/bigchaindb/issues/797 is not implemented yet. The above was meant to go around the limitation of having to pass lists. 

Did you mean to say that if a `tuple` is passed it will cause troubles?

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+
+@_prepare_transaction.register(CreateOperation)
+def _prepare_create_transaction_dispatcher(operation, **kwargs):
+    del kwargs['inputs']
+    return prepare_create_transaction(**kwargs)
+
+
+@_prepare_transaction.register(TransferOperation)
+def _prepare_transfer_transaction_dispatcher(operation, **kwargs):
+    del kwargs['owners_before']
```
sbellem: Yes, the registered function will receive all the arguments of the generic function (the one decorated with `@singledispatch`). The arguments that are not needed are simply deleted.

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,229 @@
...
+    transaction = Transaction.transfer(
+        fulfillments,
+        owners_after,
+        asset,
+        metadata=metadata,
+    )
+    return transaction.to_dict()
+
+
+def fulfill_transaction(transaction, *, private_keys):
```
r-marques: Do we really need the `*`? There are no keyword elements in this method

bigchaindb_driver/driver.py
```diff
@@ -183,6 +232,44 @@ def status(self, txid):
         path = self.path + txid + '/status'
         return self.transport.forward_request(method='GET', path=path)
 
+    def create(self, asset=None, verifying_key=None, signing_key=None):
```
r-marques: Do we still need the `create` and `transfer` methods?

bigchaindb_driver/offchain.py
r-marques: `Transaction.sign` already takes care of both cases

bigchaindb_driver/offchain.py
r-marques: At this point `input_` is actually a condition. So here you are actually creating the fulfillments from the conditions

bigchaindb_driver/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        asset = Asset.from_dict(asset)
+    else:
+        try:
+            asset = Asset(**asset)
+        except (AttributeError, TypeError):
+            if not asset:
+                asset = None
```
r-marques: This is ok for the scope of this pull request.
But overall there is still a bit of confusion because some parts of the code try to be isolated from common (argument validation, raising of exceptions) while other parts still rely on common to perform these tasks

bigchaindb_driver/offchain.py
r-marques: yes

sbellem: yes, you are right I think, maybe too much wishful thinking on my part here -- will review and adjust accordingly

bigchaindb_driver/offchain.py
sbellem: as discussed offline, the field name (key) is called `owners_before` but its value is the `owners_after` from the condition

just to make this clearer for others, say you have a transaction, `tx` you wish to consume:

```python
cid = 0
condition = tx['transaction']['conditions'][cid]
input_ = {
    'fulfillment': condition['condition']['details'],
    'input': {
        'cid': cid,
        'txid': tx['id'],
    },
    'owners_before': condition['owners_after'],
}
```

bigchaindb_driver/offchain.py
sbellem: As discussed offline, the reason for doing so was that this will allow us to introduce the support for hashlocks and other cryptoconditions goodies in a smooth way.

Hashlocks won't require private keys, and therefore we can add a default value of `None` without breaking current users code.

bigchaindb_driver/driver.py
sbellem: As discussed offline, we do not. We will add a deprecation notice for this version and we'll remove it in the subsequent version.

bigchaindb_driver/offchain.py
diminator: sure, I understand the deletion, but at the time not `owners_before`.
This has been resolved in one of your other comments

bigchaindb_driver/offchain.py
diminator: ok - maybe then a line of comment in the code?

bigchaindb_driver/offchain.py
diminator: looking at `_normalize_asset` it seems that the difference if very subtle (haven't figured it out yet)

bigchaindb_driver/utils.py
diminator: It is quite tedious to see the difference between

`asset = Asset.from_dict(asset)`
and
`asset = Asset(**asset)`

Whats the structure and type of `asset` in both cases? can you document/comment in the code pls

bigchaindb_driver/offchain.py
diminator: I assume that `input_` is a transaction since it has the key `fulfillment`?
really unclear what is happening here

bigchaindb_driver/offchain.py
sbellem: @diminator Added admonitions `.. important::` e.g.: https://github.com/bigchaindb/bigchaindb-driver/pull/143/files#diff-0de47fef8ba327e59600c5d4d879e81eR157

bigchaindb_driver/offchain.py
sbellem: yes, agreed, will document the `inputs` parameter, i.e.: give an example

bigchaindb_driver/offchain.py
sbellem: Yes, this is specific to how `common.transaction.Asset` works.

bigchaindb_driver/utils.py
sbellem: Ok, yes, will do so.

sbellem:
Note to @diminator and @r-marques: all comments have been addressed except for those that have  `todo` notes (in https://github.com/bigchaindb/bigchaindb-driver/pull/143/commits/87f984fbba00a7d16450641adcb90baae030b5a1)

Will write up the required docs, and will let you know once done.

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,325 @@
...
+        asset (:obj:`dict`, optional): The asset being created. Defaults
+            to ``None``.
+        metadata (:obj:`dict`, optional): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"CREATE"`` transaction.
+
+    .. important:: If ``owners_after`` is not given, or evaluates to
+        ``False``, it will be set equal to ``woners_before``::
```
sohkai: ``owners_before``

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,325 @@
...
+        owners_after (:obj:`list` | :obj:`tuple` | :obj:`str`, optional):
+            One or more public keys representing the new owner(s) of the
+            asset being created. Defaults to ``None``.
+        asset (:obj:`dict`, optional): The asset being created. Defaults
+            to ``None``.
+        metadata (:obj:`dict`, optional): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"CREATE"`` transaction.
```
sohkai: We usually refer to `CREATE` and `TRANSFER` transactions without the quotes (e.g. in the old docs: https://docs.bigchaindb.com/projects/server/en/v0.6.0/drivers-clients/python-server-api-examples.html).

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,325 @@
...
+                 'type': 'fulfillment',
+                 'type_id': 4},
+                'input': None,
+                'owners_before': ['3Cxh1eKZk3Wp9KGBWFS7iVde465UvqUKnEqTg2M']}],
+              'metadata': None,
+              'operation': 'CREATE',
+              'timestamp': '1479393278'},
+             'version': 1}
+
+        Then, the input may be constrcuted in this way::
```
sohkai: constr**uc**ted

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,325 @@
...
+            more public keys representing the new owner(s) of the asset
+            being transferred.
+        asset (:obj:`dict`): The asset being transferred.
+        metadata (:obj:`dict`): Metadata associated with the
+            transaction. Defaults to ``None``.
+
+    Returns:
+        dict: The prepared ``"TRANSFER"`` transaction.
+
+    Example:
```
sohkai: I would maybe put this in one of the docs files and link to it from here. This might be a little bit excessive for docstrings, especially for people who are going to be changing this file.

bigchaindb_driver/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+class CreateOperation:
+    """Class representing the ``'CREATE'`` transaction operation."""
+
+
+class TransferOperation:
+    """Class representing the ``'TRANSFER'`` transaction operation."""
+
+
+ops_map = {
```
sohkai: Use an Enum class instead, and just use values from that instead of strings for the two types of transactions. This should eventually land somewhere upstream.

Single dispatch is still possible with [`@valuedispatch`](http://lukasz.langa.pl/8/single-dispatch-generic-functions/#generic-history).

sbellem:
@diminator @r-marques done.

Let me know if you have any more comments, feedback, questions.

I have addressed your feedback in new commits, to make it easier to review.

Once you are done, I will rebase the branch to clean up the history.

bigchaindb_driver/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        operation = ops_map[operation]()
+    except KeyError:
+        pass
+
+    return operation
+
+
+def _normalize_asset(asset, is_transfer=False):
```
sohkai: Might as well be `@single_dispatch`ed.

bigchaindb_driver/driver.py
```diff
@@ -119,43 +120,100 @@ class TransactionsEndpoint(NamespacedDriver):
...
+
+        .. important::
+
+            **CREATE operations**
+
+            * ``owners_before`` MUST be set.
+            * ``owners_after``, ``asset``, and ``metadata`` MAY be set.
+            * The argument ``inputs`` is ignored.
+            * If ``owners_after`` is not given, or evaluates to
+              ``False``, it will be set equal to ``woners_before``::
```
sohkai: ``owners_before``

bigchaindb_driver/offchain.py
```diff
@@ -0,0 +1,325 @@
...
+    # NOTE: Needed for the time being. See
+    # https://github.com/bigchaindb/bigchaindb/issues/797
+    if isinstance(private_keys, tuple):
+        private_keys = list(private_keys)
+
+    transaction_obj = Transaction.from_dict(transaction)
+    try:
+        signed_transaction = transaction_obj.sign(private_keys)
+    except KeypairMismatchException as exc:
+        raise MissingSigningKeyError('A signing key is missing!') from exc
```
sohkai: Maybe we should note this exception in the docstrings both here and upstream in the driver's static method?

docs/advanced-usage.rst
```diff
@@ -144,66 +183,146 @@ index ``cid``.
...
+
+    signed_tx_transfer = bdb.transactions.fulfill(
+        tx_transfer,
+        private_keys=alice.signing_key,
+    )
+
+If you look at the content of ``signed_tx_transfer`` you should see the added
+fulfilment uri, holding the signature:
+
+.. code-block:: python
```
sohkai: I would just put the `signed_tx_transfer['transaction']['fulfillments'][0]['fulfillment']` block here.

docs/advanced-usage.rst
```diff
@@ -144,66 +183,146 @@ index ``cid``.
...
+
+More precisely:
+
+.. code-block:: python
+
+    >>> signed_tx_transfer['transaction']['fulfillments'][0]['fulfillment']
+    >>> 'cf:4:IMe7QSL5xRAYIlXon76ZonWktR0NI02M8rAG1bN-ugg4S_S7Obu7E-HtL2ZjM3tcKKfoaspMhyx17Eg2KBijylZMxv1NvAD0j8uJP1WOb2AP6ezJorcw6TA5n-cmuwkE'
+
+We have yet to send the transaction over to a BigchainDB node, as both
+preparing and fulfilling a transaction are done "offchain", that is withou the
```
sohkai: withou**t**

docs/advanced-usage.rst
```diff
@@ -256,48 +376,144 @@ Multiple Owners
...
+        'weight': 1}],
+      'threshold': 2,
+      'type': 'fulfillment',
+      'type_id': 2},
+     'input': {'cid': 0,
+      'txid': '3128572d430f9ae23010f11e444d1d12e35f90095cf5f0da98acfcada5eed0fe'},
+     'owners_before': ['3Cxh1eKZk3Wp9KGBWFS7iVde465UvqUKnEqTg2MW4wNf',
+      'EcRawy3Y22eAUSS94vLF8BVJi62wbqbD9iSUSUNU9wAA']}
+
+Them we need the asset:
```
sohkai: The**n**

tests/test_driver.py
```diff
@@ -133,3 +136,41 @@ def test_status_not_found(self, driver):
         txid = 'dummy_id'
         with raises(NotFoundError):
             driver.transactions.status(txid)
+
+    def test_prepare(self, driver, alice_pubkey):
+        transaction = driver.transactions.prepare(owners_before=[alice_pubkey])
```
sohkai: Maybe this could be a fixture that can then also be used by `test_fulfill` rather than the hardcoded `unsigned_transaction`?

tests/test_offchain.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_prepare_transaction_raises():
+    from bigchaindb_driver.offchain import prepare_transaction
+    from bigchaindb_driver.exceptions import BigchaindbException
+    with raises(BigchaindbException):
+        prepare_transaction(operation=None)
+
+
+@mark.parametrize('asset', (
```
sohkai: Might as well parametrize some `metadata` too?

tests/test_offchain.py
```diff
@@ -0,0 +1,91 @@
...
+    None, {}, '', {'data': {'msg': 'Hello BigchainDB!'}},
+))
+def test_prepare_create_transaction(asset, alice_keypair):
+    from bigchaindb_driver.offchain import prepare_create_transaction
+    create_transaction = prepare_create_transaction(
+        owners_before=alice_keypair.vk, asset=asset)
+    assert 'id' in create_transaction
+    assert 'transaction' in create_transaction
+    assert 'version' in create_transaction
+    assert 'asset' in create_transaction['transaction']
```
sohkai: Would be nice to actually check the `asset` matched up with the given one. (Or checking assets and metadata can be separate tests)

tests/test_offchain.py
```diff
@@ -0,0 +1,91 @@
...
+        },
+        'owners_before': condition['owners_after']
+    }
+    asset = alice_transaction['transaction']['asset']
+    transfer_transaction = prepare_transfer_transaction(
+        inputs=input_, owners_after=bob_keypair.vk, asset=asset)
+    assert 'id' in transfer_transaction
+    assert 'transaction' in transfer_transaction
+    assert 'version' in transfer_transaction
+    assert 'asset' in transfer_transaction['transaction']
```
sohkai: Would be nice to double check if the asset generated for the transfer matches up with the input's.

bigchaindb_driver/offchain.py
sbellem: The reason is I used the quotes was to keep a connection with the parameter it is linked to, i.e.: `operation`.

The potential user of the library reading the docs, will pass `'CREATE'` or `'TRANSFER'`, with quotes.

That is the reasoning behind the choice to use quotes.

bigchaindb_driver/offchain.py
sbellem: Yes. Let's see what @diminator says here, because he had asked about an example of input.

bigchaindb_driver/offchain.py
sbellem: yes

sbellem: done

tests/test_driver.py
sbellem: the hard-coding was intentional as it provides a view of what the payload looks like

PR #146, opened 2016-11-22
Remove user provided transaction timestamp [untested]

libscott:
A PR to remove user provided timestamp from tests and docuementation.

This has not been tested since I got some errors running `docker-compose run --rm bdb-driver pytest -v`, about (E: Failed to fetch http://deb.debian.org/debian/pool/main/g/gpm/libgpm2_1.20.4-6.1+b2_amd64.deb  Could not resolve 'deb.debian.org').


sohkai:
We should probably wait to merge this until we mark a driver release that supports BDB 0.7. Once that's done, we can say everything afterwards is targeting latest BDB.

r-marques:
cherry picked this commit in https://github.com/bigchaindb/bigchaindb-driver/pull/158/commits/0c67490c4b43725ab1b223e2d75169d9e062f117 and closing this in favor of #158

PR #150, opened 2016-11-23
Update advanced usage docs

sbellem:


ttmc:
Maybe change the title of the page from **Advanced Usage** to **Advanced Usage Examples** so it's consistent with the previous page title (**Basic Usage Examples**). The filename doesn't have to be changed.

ttmc:
This sentence is out of date:

"In BigchainDB, only the federation nodes are allowed to create digital assets, by doing a special kind of transaction: a `CREATE` transaction."

The example code below that sentence is more accurate. It says:
```python
tx = bdb.transactions.prepare(operation='CREATE',
                              owners_before=alice.verifying_key,
                              asset=digital_asset_payload)

# All transactions need to be signed by the user creating the transaction.
signed_tx = bdb.transactions.fulfill(tx, private_keys=alice.signing_key)
```

I think it's also possible for a _set_ of multiple "owners before" to sign a CREATE transaction. Maybe this isn't the place to bring that up, but at least we shouldn't say the _only_ option is a single creator.

docs/advanced-usage.rst
```diff
@@ -216,89 +335,197 @@ If we try to create another transaction with the same input as before, the
...
+        print(e, e.__cause__, sep='\n')
+
+You should see a message similar to:
+
+.. code-block:: text
+
+    A signing key is missing!
+    Public key EcRawy3Y22eAUSS94vLF8BVJi62wbqbD9iSUSUNU9wAA is not a pair to any of the private keys
+
+Notice that is ``bob``'s public key in the above message:
```
sohkai: Notice that `bob`'s public key is the one in the above message:

docs/advanced-usage.rst
```diff
@@ -216,89 +335,197 @@ If we try to create another transaction with the same input as before, the
...
+
+.. code-block:: python
+
+    >>> bob.verifying_key
+    'EcRawy3Y22eAUSS94vLF8BVJi62wbqbD9iSUSUNU9wAA'
+
+Same goes for ``alice``. Try it!
```
sohkai: **And the same**

docs/advanced-usage.rst
```diff
@@ -216,89 +335,197 @@ If we try to create another transaction with the same input as before, the
...
+
+.. code-block:: python
+
+    >>> bdb.transactions.status(sent_car_transfer_tx['id'])
+    {'status': 'valid'}
+
+Done!
+
+Happy, ``alice`` and ``bob`` have successfully transferred the ownership of
+their car to ``carol``, and go on exploring the countless galaxies of the
```
sohkai: and **can** go

ttmc:
In the first example, i.e.
```python
tx = bdb.transactions.prepare(operation='CREATE',
                              owners_before=alice.verifying_key,
                              asset=digital_asset_payload)
```

I wonder if it's possible to put metadata in the transaction as well. These are "advanced" usage examples, after all. We got a question about how to include metadata last week (via Gitter).

sbellem:
> I think it's also possible for a set of multiple "owners before" to sign a CREATE transaction. Maybe this isn't the place to bring that up, but at least we shouldn't say the only option is a single creator.

version 0.7.0 of BigchainDB does not support it, see https://github.com/bigchaindb/bigchaindb/blob/v0.7.0/bigchaindb/common/transaction.py#L689-L691

docs/advanced-usage.rst
```diff
@@ -216,89 +335,197 @@ If we try to create another transaction with the same input as before, the
...
+
+if ``alice`` and ``bob`` wish to check the status of the transfer they may use
+the :meth:`~bigchaindb_driver.BigchainDB.transactions.status` endpoint:
+
+.. code-block:: python
+
+    >>> bdb.transactions.status(sent_car_transfer_tx['id'])
+    {'status': 'valid'}
+
+Done!
```
sbellem: Why? Isn't this very subjective to the reader?

docs/advanced-usage.rst
```diff
@@ -216,89 +335,197 @@ If we try to create another transaction with the same input as before, the
...
+
+Happy, ``alice`` and ``bob`` have successfully transferred the ownership of
```
sbellem: Why? Again, here, isn't this subjective?

sbellem:
> I wonder if it's possible to put metadata in the transaction as well. These are "advanced" usage examples, after all. We got a question about how to include metadata last week (via Gitter).

yes, good point

How about having a dedicated section?

ttmc:
About showing how to include metadata in the transaction: I was looking at the updated **Basic Usage Examples** in PR #155 and saw that those now show how to include metadata. It make sense to show that there --- including metadata isn't complicated, after all. It's basic usage.

Therefore it's probably not necessary to show how to include metadata in this page.

sbellem:
> About showing how to include metadata in the transaction: I was looking at the updated Basic Usage Examples in PR #155 and saw that those now show how to include metadata. It make sense to show that there --- including metadata isn't complicated, after all. It's basic usage.
> 
> Therefore it's probably not necessary to show how to include metadata in this page.

Yes, ok!

docs/advanced-usage.rst
```diff
@@ -144,66 +182,146 @@ index ``cid``.
...
+    input_ = {
+        'fulfillment': condition['condition']['details'],
+        'input': {
+            'cid': cid,
+            'txid': tx['id'],
+        },
+        'owners_before': condition['owners_after'],
+    }
+
+The asset, can be directly retrieved from the input ``tx``:
```
TimDaub: I think it's good that we describe what's happening here, but ultimately I think we should allow the user to have some functionality that is similar to `Transaction.to_input` as it's quite painful to construct this payload by oneself (+ not everybody will always get these things right).

docs/advanced-usage.rst
```diff
@@ -216,89 +334,197 @@ If we try to create another transaction with the same input as before, the
...
+
+The asset can be transfered as soon as each of the ``owners_after`` fulfills
+the transaction, that is ``alice`` and ``bob``.
+
+To do so, simply provide a list of all private keys to the fulfill method:
+
+.. code-block:: python
+
+    signed_car_transfer_tx = bdb.transactions.fulfill(
+        car_transfer_tx, private_keys=[alice.signing_key, bob.signing_key]
```
vrde: This is a workaround for now. In real life you won't have two `private_keys` of two different users loaded in your memory. This use case will be resolved by having Alice signing and passing the transaction to Bob, so he can sign it as well, but this will happen *off–chain*.

I think this example is misleading, I'd be more explicit saying that this will change once we have a better system to have a partial transaction that can be passed around and signed.

docs/advanced-usage.rst
sbellem: The current design of the driver is aiming to be a low-level layer, in which users are expected to have the know-how, or to go through a learning curve of learning how to construct the payloads on their own.

Once this low-level layer is solid enough we definitely wish to consider adding an additional layer that does provide an easy-to-use interface to compose crypto-conditions, and subsequently transactions.

I understand that the above may be debatable or questionable. I just created an issue which aims to address the underlying problem here, which is a need for clarification as to what is expected from the current driver.

The issue is at https://github.com/bigchaindb/bigchaindb-driver/issues/156

The current driver is the result of important design decisions, which unfortunately have yet to be documented, hence the above issue. It is hoped that once documented, multiple things will be clearer.

docs/advanced-usage.rst
sbellem: Well, at this point in time the driver is limited by `bigchaindb.common`, see issue https://github.com/bigchaindb/bigchaindb/issues/729

We can definitely add a note indicating our vision, and intentions on the matter.

docs/advanced-usage.rst
sohkai: It certainly is subjective; I'm just sayin' it feels out of place for me.

docs/advanced-usage.rst
sohkai: It just reads a little odd, but yes, generally criticism on writing is subjective.

PR #155, opened 2016-11-23
Update basic usage docs

### LINKED ISSUE #151: Update basic usage example docs
replace `.create` and `.transfer` with (`.prepare`, `.fulfill`, `.send`)

sbellem:
closes #151

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -94,6 +94,17 @@ We'll suppose that the bike belongs to Alice, and that it will be transferred
 to Bob.
 
 
+Metadata Definition (*optional*)
+--------------------------------
+You can `optionally` add metadata to a transaction. Any dictionary is accepted.
+
+For example:
+
+.. code-block:: python
+
+    metadata = {'planet': 'earth'}
+
+
 Cryptographic Identities Generation
 -----------------------------------
 Alice, and Bob are represented by signing/verifying key pairs. The signing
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -110,53 +110,112 @@ who claims to be the signee.
...
-                                          signing_key=alice.signing_key,
-                                          asset=bicycle)
+   prepared_creation_tx = bdb.transactions.prepare(
+        operation='CREATE',
+        owners_before=alice.verifying_key,
+        asset=bicycle,
+        metadata=metadata,
+   )
 
 The ``creation_tx`` dictionary should be similar to:
```
ttmc: `creation_tx` changed its name to `prepared_creation_tx`

ttmc:
I noticed that some of the Python command lines begin with `>>>` and others begin with nothing. I prefer the nothing version because it's easier to copy & paste, but regardless of which version you prefer, you should stick with one or the other.

ttmc:
I have no more comments to make about this pull request (at this time).

sbellem:
> I noticed that some of the Python command lines begin with >>> and others begin with nothing. You should stick with one or the other.

Indeed!

Was not so sure on to handle this.

1. Making the code easy to copy and paste: YES!
2. Making it clear what is the command, versus what is the output: YES!

1 and 2 conflict :confused:

PR #157, opened 2016-11-24
Properly set default connection node if none given

sohkai:
Oops, turns out we didn't actually use the default node.

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -45,7 +45,7 @@ class BigchainDB:
         self._nodes = nodes if nodes else (DEFAULT_NODE,)
         self._verifying_key = verifying_key
         self._signing_key = signing_key
-        self._transport = transport_class(*self.nodes)
+        self._transport = transport_class(*self._nodes)
         self._transactions = TransactionsEndpoint(self)
 
     @property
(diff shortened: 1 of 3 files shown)


bigchaindb_driver/driver.py
```diff
@@ -45,7 +45,7 @@ def __init__(self,
         self._nodes = nodes if nodes else (DEFAULT_NODE,)
         self._verifying_key = verifying_key
         self._signing_key = signing_key
-        self._transport = transport_class(*nodes)
+        self._transport = transport_class(*self.nodes)
```
sbellem: We can simply use `self._nodes` here, no need to go through the property, no?

docker-compose.yml
```diff
@@ -17,13 +17,13 @@ services:
...
+      - "28015:28015"
   bdb-server:
     build: ./compose/server
     environment:
       BIGCHAINDB_DATABASE_HOST: rdb
       BIGCHAINDB_API_ENDPOINT: http://bdb-server:9984/api/v1
       BIGCHAINDB_SERVER_BIND: 0.0.0.0:9984
     ports:
-      - "9984"
+      - "32675:9984"
```
sbellem: not pinning the ports was intentional, since when working with the server there may be clashes, hence not pinning is the safest to avoid clashes

sohkai: Oops, must've been an artifact from testing. Fixed in 92251f1.

bigchaindb_driver/driver.py
sohkai: Fixed in 852eabb.

sbellem:
@sohkai all good! :+1: 

Could you rebase, or squash? This can be just one commit, right?

PR #158, opened 2016-11-24
Feat/134/support divisible assets

r-marques:
Since the driver works mostly with dicts it required no changes to support divisible assets.
This pr is now mostly to support the latest changes in the bigchaindb server.


sbellem:
:+1:

PR #167, opened 2016-11-25
Compatibility matrix

sbellem:


ttmc:
:+1:

README.rst
```diff
@@ -27,8 +27,22 @@ BigchainDB Python Driver
...
+
+Compatibility Matrix
+--------------------
+
++-----------------------+-----------------------+
+| **BigchainDB Server** | **BigchainDB Driver** |
++=======================+=======================+
+| ``0.7.x``             | ``0.0.3``             |
++-----------------------+-----------------------+
+| ``master``            | ``master``            |
```
ttmc: @sbellem I realize you already merged this pull request, but it occurred to me that it might be overly optimistic to claim that the two `master` branches are always compatible with each other. Usually what happens is the server gets ahead of the Python driver, and then the Python driver catches up.

sbellem: > it might be overly optimistic to claim that the two master branches are always compatible with each other. Usually what happens is the server gets ahead of the Python driver, and then the Python driver catches up.

Yes, I agree.

It was nevertheless intentional on my part, for the time being to have it this way for a few reasons:

1. It is currently true.
2. In the context of the [DappHack]( http://www.dapphack.org/) event (DApp Hackathon and unConference about decentralised systems), and of our current users, I felt it was better to keep things as simple as possible.
3. I wanted to put a note about how there may be delays for the synchronization to occur, but was hesitant about the wording, and eventually felt it would be simpler to put no note for now, and to add one once it applies. In other words, I thought that if this row of the matrix could be dynamic that would be better.

That being said, what do you think about putting a simple and clear note explaining in a concise way that although we do our best to keep them in sync, there may be occasional delays?

If the HTTP server API contract is defined before the implementation, it should be possible to keep them in sync though, and even to have the driver(s) waiting for the server implementation.

ttmc: Yes, it would be worthwhile to note that the two master branches sometimes get out of sync.

sbellem: https://github.com/bigchaindb/bigchaindb-driver/pull/173

PR #172, opened 2016-11-26
Add hacking file and doc section

sbellem:
Due to the current dependency on bigchaindb server code, when the master branch is needed installing with pip requires using the flag `--process-dependency-links`, which may be unknown to many users. The added docs aim to provide the possibly missing information.


ttmc:
These are instructions to use `pip` to install from the latest `master` branch, so why not put them in a new section in the existing **Quickstart / Installation** page?

ttmc:
:+1:

PR #176, opened 2016-11-29
Add handcrafting transaction docs rough cut

sbellem:
**Note to reviewers**:

* The plan of this PR is to provide a rough cut of working examples that have almost no dependence on the driver itself.
* The examples are meant to be compatible with the latest PyPI release, 0.1.2.
* Once merged, the plan is to make a maintenance release (0.1.3) for which these new docs will be available.

Subsequent work will be part of 0.2.0 such that the driver will be brought up to speed with the latest server code.

As for the examples, work will be done to remove all dependencies on the driver, such that they can be shown under the server docs,

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -1,3 +1,5 @@
+.. _advanced-usage:
+
 ***********************
 Advanced Usage Examples
 ***********************
--- a/docs/handcraft.rst
+++ b/docs/handcraft.rst
@@ -1,0 +1,2043 @@
+#########################
+Handcrafting Transactions
+#########################
+
+For those who wish to assemble the transaction payload "by hand".
+
+.. contents::
+    :local:
+    :depth: 1
+
+********
+Overview
+********
+
+Submitting a transaction to a BigchainDB node consists of three main steps:
+
+1. Preparing the transaction payload.
+2. Fulfilling the prepared transaction payload.
+3. Sending the transaction payload via HTTPS.
+
+Step 1 and 2 can be performed offline on the client side. That is, they do not
+require any connection to any BigchainDB node.
+
+For convenience's sake, some utilites are provided to prepare and fulfill a
+transaction via the :class:`~.bigchaindb_driver.BigchainDB` class, and via the
+:mod:`~bigchaindb_driver.offchain` module.
+
+For an introduction on using these utilities, see the :ref:`basic-usage` or
+:ref:`advanced-usage` sections.
+
+The rest of this document will guide you through completing steps 1 and 2
+yourself by revisiting some of the examples provided in the usage sections.
+We will:
+
+* provide all values, including the default ones;
+* generate the transaction id;
+* learn to use crypto-conditions to generate a condition that locks the
+  transaction, hence protecting it from being consumed by an unauthorized user;
+* learn to use crypto-conditions to generate a fulfillment that unlocks
+  the transaction asset, and consequently enact the ownership transfer.
+
+In order to perform all of the above, we'll need help from a few libraries:
+
+* :mod:`json`: to serialize the transaction dictionary into a JSON formatted
+  string.
+* `sha3`_: to hash the serialized transaction
+* `cryptoconditions`_: to create conditions and fulfillments
+
+
+High-level view of the transaction payload
+==========================================
+For detailled documentation on the transaction model/schema, please consult
+`The Transaction Model`_ and `The Transaction Schema`_ documentation.
+
+Here is just a quick survey, highlighting key points.
+
+From the point of view of Python, The transaction payload is simply a
+dictionary with a bit of nested structure.
+
+The first level has three keys;
+
+* ``id`` -- a ``str``
+* ``version`` -- an ``int``
+* ``transaction`` -- a ``dict``
+
+Because a transaction is signed, the ``id`` needs to be provided on the client
+side.
+
+When you assemble the payload you'll have:
+
+.. ipython::
+
+    payload = {
+        'version': 1,
+        'transaction': {...},
+    }
+
+and you'll generate the ``id`` by hashing the above.
+
+
+.. important:: **Implications of having Signed Payloads**
+
+    Because transactions are signed by the client submitting them, various
+    values that could otherwise be generated on the server side need to be
+    generated on the client side.
+
+    These values are:
+
+    * **transaction id**, which is a hash of the entire payload, without the
+      signature(s)
+    * **asset id**
+    * **metadata id**
+    * any optional value, such as ``version`` which defaults to ``1``
+
+    This makes the assembling of a payload more involved as one needs to
+    provide all values regardless of whether there are defaults or not.
+
+
+The transaction body
+--------------------
+The transaction body is made up of the following keys:
+
+* ``asset`` -- ``dict``
+* ``metadata`` -- ``dict``
+* ``operation`` -- ``str``
+* ``conditions`` -- ``list`` of ``dict``
+* ``fulfillments`` -- ``list`` of ``dict``
+
+asset
+^^^^^
+
+.. code-block:: python
+
+    asset = {
+        'data': {},
+        'divisible': False,
+        'refillable': False,
+        'updatable': False,
+        'id': '',
+    }
+
+Example of an asset payload:
+
+.. code-block:: python
+
+    asset = {
+        'data': {
+            'bicycle': {
+                'manufacturer': 'bkfab',
+                'serial_number': 'abcd1234',
+            },
+        },
+        'divisible': False,
+        'refillable': False,
+        'updatable': False,
+        'id': '7ab63c48-4c24-41df-a1bd-934bb609a7f7',
+    }
+
+.. note:: In a many client-server architectures, the values for the keys:
+
+    * ``'divisible'``
+    * ``'refillable'``
+    * ``'updatable'``
+    * ``'id'``
+
+   could all be generated on the server side.
+
+   `In the case of BigchainDB, because we rely on cryptographic signatures,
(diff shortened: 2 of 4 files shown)


dev.yml
```diff
@@ -0,0 +1,32 @@
+version: '2'
```
sohkai: Can this be a different PR?

docs/advanced-usage.rst
```diff
@@ -1,3 +1,5 @@
+.. _advanced-usage:
```
sohkai: I assume that this is basically just #187?

docs/handcraft.rst
```diff
@@ -0,0 +1,1940 @@
...
+
+.. contents::
+    :local:
+    :depth: 1
+
+********
+Overview
+********
+
+Submitting a transaction to a BigchainDB node consist of three main steps:
```
sohkai: consist*s*

docs/handcraft.rst
```diff
@@ -0,0 +1,1940 @@
...
+
+1. Preparing the transaction payload.
+2. Fulfilling the prepared transaction payload.
+3. Sending the transaction payload via HTTPS.
+
+Step 1 and 2 can be performed offline on the client side. That is, they do not
+require any connection to any BigchainDB node.
+
+For convenience's sake some utilites are provided to prepare and fulfill a
```
sohkai: sake**,**

docs/handcraft.rst
```diff
@@ -0,0 +1,1940 @@
...
+transaction via the :class:`~.bigchaindb_driver.BigchainDB` class, and via the
+:mod:`~bigchaindb_driver.offchain` module.
```
sohkai: I'd reformulate the next two paragraphs and do some merging. It'd be:

> via the :mod:`offchain` module. **For an introduction on using these utilities, see the :ref:`basic-usage` or :ref:`advanced-usage` sections.**
> 
> **The rest of this document will guide you through completing steps 1 and 2 yourself by revisiting some of the examples provided in the usage sections. We will:**..

dev.yml
sbellem: yes, sorry: https://github.com/bigchaindb/bigchaindb-driver/pull/190

docs/advanced-usage.rst
sbellem: It was added because it is referenced to in `handcraft.rst`.

docs/handcraft.rst
sbellem: adressed in https://github.com/bigchaindb/bigchaindb-driver/pull/176/commits/4cf6b191bddf1479bfcae98301a194fabd3929f2

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+
+fulfillments
+^^^^^^^^^^^^
+At the moment, we have to distinguish between fulfillment payloads **before**
+fulfillment and **after** fulfillment. For signature-based schemes, this means,
+before signature(s), and after signature(s).
+
+Moreover payloads for ``CREATE`` operations are a bit different.
+
+.. note:: We hope to be able to simplify the payload strucutre and validation,
```
r-marques: typo `strucuture`

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+            },
+            'uri': 'cc:4:20:bOZjTedaOgPsbYjh3QeOEQCj1o1lIvVefR71sS8egnM:96'
+        },
+        'owners_after': ['8L6ngTZ5ixuFEr1GiunrFNWtGkft4swWWArXjWJu2Uwc'],
+    }
+    
+
+fulfillments
+^^^^^^^^^^^^
+At the moment, we have to distinguish between fulfillment payloads **before**
```
r-marques: I find this paragraph a bit confusing. I understand what you are trying to say.
What actually want is to construct the fulfillment object but without the signature.

One reason to do this is to avoid exposure of the private key. I may want to use some utility to generate the payloads (e.g. `bigchaindb_driver`, `bdb-cli`) and then I want to sign the paylod if I feel everything is ok. Possibly on another machine where I store my private keys

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+
+    In [0]: alice.verifying_key
+
+We are now going to craft this payload by hand.
+
+Extract asset id and metadata id:
+
+.. ipython::
+
+    In [0]: asset_id = prepared_creation_tx['transaction']['asset']['id']
```
r-marques: instead of doing this , maybe `str(uuid.uuid4())`.

Unless you want to end up with the exact same payload which is fine, but then just add a note about it. Or else it feels like cheating first creating a transaction with the driver and then using some of its values

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+       ...:             'serial_number': 'abcd1234',
+       ...:         },
+       ...:     },
+       ...:     'divisible': False,
+       ...:     'refillable': False,
+       ...:     'updatable': False,
+       ...:     'id': asset_id,
+       ...: }
+
+.. note:: In a signature-less system, the values for the keys:
```
r-marques: I am not sure that this is worth mentioning since this is not going to happen. And its also bad practice because then you would have to trust the server not to change the payload

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+.. ipython::
+
+    In [0]: condition = {
+       ...:     'amount': 1,
+       ...:     'cid': 0,
+       ...:     'condition': {
+       ...:         'details': ed25519.to_dict(),
+       ...:         'uri': ed25519.condition_uri,
+       ...:     },
+       ...:     'owners_after': (alice.verifying_key,),
```
r-marques: Does the serialize json replaces tuples with lists?

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+
+    In [0]: conditions = (condition,)
+
+The key part is the condition uri:
+
+.. ipython::
+
+    In [0]: ed25519.condition_uri
+
+.. todo:: Explain the basics of the above. cc, 4, 20, the hash of the key, 96
```
r-marques: You can just point out that its part of the `cc` spec and point them to the ietf draft

docs/handcraft.rst
sbellem: Ah ok. The goal was to outline some key differences for developers who are familiar with HTTP server - client architectures in which the payloads can be very small since the server takes care of filling in default values.

But it can be removed.

docs/handcraft.rst
sbellem: Indeed! How cool is that?!

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+
+    In [0]: prepared_creation_tx_bk = deepcopy(prepared_creation_tx)
+
+    In [0]: # set fulfillment to None
+
+    In [0]: prepared_creation_tx['transaction']['fulfillments'][0]['fulfillment'] = None
+ 
+    In [0]: handcrafted_creation_tx == prepared_creation_tx
+
+Still not equal because we used tuples instead of strings
```
r-marques: lists?

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+    import cryptoconditions
+    
+    from bigchaindb_driver.crypto import generate_keypair
+
+
+    alice = generate_keypair()
+
+    operation = 'CREATE'
+    
+    asset_id = str(uuid4())
```
r-marques: Here it is. You may ignore my previous comment on the `uuid`

docs/handcraft.rst
sbellem: Yes, this is more like scaffolding that could be removed. What initially happened is that I pretty much reversed-engineered common, and wanted to double check that the hashes, and signatures were the same, hence re-using the asset id.

So what do you think, we could probably remove altogether the comparisons with the driver offchain tools (`prepare_transaction`, `fulfill_transaction`)?

docs/handcraft.rst
```diff
@@ -0,0 +1,2049 @@
...
+
+    In [0]: prepared_transfer_tx_bk = deepcopy(prepared_transfer_tx)
+
+    In [0]: # set fulfillment to None
+
+    In [0]: prepared_transfer_tx['transaction']['fulfillments'][0]['fulfillment'] = None
+
+    In [0]: handcrafted_transfer_tx == prepared_transfer_tx
+
+Still not equal because we used tuples instead of strings
```
r-marques: lists?

docs/handcraft.rst
sbellem: Ok, should I remove the entire comment? I do not mind. As far as the entire page is concerned, what really matter are the examples.

docs/handcraft.rst
sbellem: Well, yes, here because I did not care about comparing with the payloads generated by `prepare_transaction` and `fulfill_transaction`

docs/handcraft.rst
r-marques: No. After reading the entire document I like it this way. First you craft a transaction and compare it with what the driver outputs. And then you craft the same transaction without the support of the driver generating the uuids yourself.

I like it because if someone is going to develop a driver they can follow the same pattern and constantly compare it with a real transaction to make sure they are doing everything ok.

docs/handcraft.rst
r-marques: Oh. Then I think is better to just say that instead of using signature-less system

r-marques:
I really like it. The language is simple and to the point and you almost make it sound simple

PR #180, opened 2016-11-30
Remove metadata uuid from usage docs

libscott:
The metadata UUID was removed upstream: https://github.com/bigchaindb/bigchaindb/blob/master/bigchaindb/common/schema/transaction.yaml#L243.

This PR addresses that change in the usage docs, however since the transactions in the documentation have IDs hardcoded, at least one of them will need to be recreated and hashed. 

Is there a protocol for doing this? I'm currently doing some work to auto generate transaction examples in the server docs for this reason.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -160,8 +160,7 @@ The ``prepared_creation_tx`` dictionary should be similar to:
          'type_id': 4},
         'input': None,
         'owners_before': ['3Cxh1eKZk3Wp9KGBWFS7iVde465UvqUKnEqTg2MW4wNf']}],
-      'metadata': {'data': {'planet': 'earth'},
-       'id': '4d406690-d6e3-48b2-ac64-9fff714f0ff3'},
+      'metadata': {'planet': 'earth'},
       'operation': 'CREATE',
      'version': 1}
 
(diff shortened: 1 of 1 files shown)


sbellem:
> Is there a protocol for doing this? I'm currently doing some work to auto generate transaction examples in the server docs for this reason.

Some of us had thought of using the [sphinx extension `autorun`](https://bitbucket.org/birkenfeld/sphinx-contrib/src/c4a1c0d34d31fc4f1a42897be1d636a8158671a5/autorun/?at=default) to auto generate outputs.

libscott:
> Some of us had thought of using the sphinx extension autorun to auto generate outputs.

Looks nice, but from what I can see, it doesn't allow you to insert `rst` markup into the output, or specify the output language apart from the input language.

sbellem:
> Looks nice, but from what I can see, it doesn't allow you to insert rst markup into the output, or specify the output language apart from the input language.

I wonder whether we need this at this point in time. I thought that all that was needed in this case was to be able to automatically generate the output of Python code.

That is why I thought that `autorun` could be useful.

sbellem:
For the time being I propose the following approach: https://github.com/bigchaindb/bigchaindb-driver/pull/185

It relies on the [IPython.sphinxext.ipython_directive](https://ipython.org/ipython-doc/3/api/generated/IPython.sphinxext.ipython_directive.html) extension, and is straightforward to use.

As for `autorun` it is currently being ported to Python 3, but not fully ready.

sbellem:
Closing as this was taken care of via https://github.com/bigchaindb/bigchaindb-driver/pull/187

PR #185, opened 2016-12-02
Docs use ipython directive

sbellem:


ttmc:
I think you'll have to add the newly-used extensions to the `docs/requirements.txt` and similarly in `setup.py`

PR #189, opened 2016-12-06
Remove master:master mapping in compatibility matrix

TimDaub:
Arguments for removing it:

- We don't have tests confirming the compatibility for every change that lands
  in master of both bigchaindb/bigchaindb and bigchaindb/bigchaindb-driver
- It involves a lot of effort to check whether or not master and master are
  compatible
- We shouldn't check manually (if we want to have this mapping, lets
  automate it)
- There has already been a case where users got confused:
  https://gitter.im/bigchaindb/bigchaindb?at=5846b0abf666c5a138d91959

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,8 +41,6 @@ Compatibility Matrix
 +=======================+=======================+
 | ``0.8.x``             | ``0.1.x``             |
 +-----------------------+-----------------------+
-| ``master``            | ``master``            |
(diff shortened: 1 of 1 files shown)


sbellem:
Ok, I guess the note underneath isn't clear enough. 

> Although we do our best to keep the master branches in sync, there may be occasional delays.

It should also be removed if we do remove the row in the matrix.

PR #196, opened 2016-12-07
Test docs building

sbellem:


docs/handcraft.rst
```diff
@@ -472,7 +472,7 @@ The ``fulfillments`` value is simply a list or a tuple of all fulfillments:
         
         In [0]: prepared_creation_tx['transaction']['fulfillments'][0]
 
-    More precisely, the value of ``'fulfillment'`:
+    More precisely, the value of ``'fulfillment'``:
 
     .. ipython::
```
r-marques: Just out of curiosity is this code actually executed in ipython when building the docs?

sbellem: I think so: https://github.com/ipython/ipython/blob/master/IPython/sphinxext/ipython_directive.py#L268

PR #203, opened 2016-12-15
Get unspents

sbellem:
**NOTE to reviewers**: 

* Some very rudimentary helper functions were added to construct simple payloads, for the tests. These helper functions were written also as a first step towards not using server code in the low level client. They are not meant to be perfect at this point in time and are meant for internal usage only.

### DIFF
--- a/bigchaindb_driver/connection.py
+++ b/bigchaindb_driver/connection.py
@@ -22,19 +22,20 @@ class Connection:
         self.node_url = node_url
         self.session = Session()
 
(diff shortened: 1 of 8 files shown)


sbellem:
> No custom code for creating transactions is necessary (especially not the functionality that was being added - common is solving it 100%).

It was added to support the tests. It is a known best practice that it is better to reduce dependencies for setting up tests. Would moving the utilities under the tests make you happy?

sbellem:
replaced by #207

PR #224, opened 2017-01-25
Feat 212 http headers

sbellem:


bigchaindb_driver/connection.py
```diff
@@ -11,31 +11,42 @@
...
 
         """
         url = self.node_url + path if path else self.node_url
         response = self.session.request(
-            method=method, url=url, json=json, params=params, **kwargs)
+            method=method,
+            url=url,
+            json=json,
+            params=params,
+            headers=headers,
```
sohkai: I assume these will override any that were added with `session.headers.update()`?

bigchaindb_driver/driver.py
```diff
@@ -31,10 +29,18 @@ def __init__(self,
                 ``'http://localhost:9984/api/v1'``.
             transport_class: Optional transport class to use.
                 Defaults to :class:`~bigchaindb_driver.transport.Transport`.
+            headers (dict): Optional headers that will be passed to the
+                ``transport_class`` instantiation, which will in turn
+                pass it to the :class:`~.connection.Connection`
+                instances. Please note that the given headers will be
+                sent on each request. To pass headers only on a
```
sohkai: I wouldn't bother with the details about how it's passed, just that headers sent this way will be sent with each request.

```python
headers (dict, optional): Headers to be pass with each request from the driver. If you'd like to pass headers only on a per-request basis...

File: compose/server/Dockerfile [hunk 3]
```diff
@@ -9,5 +9,5 @@ RUN pip install --upgrade pip ipdb ipython
 
 COPY . /usr/src/app/
 
-RUN pip install bigchaindb
+RUN pip install git+https://github.com/bigchaindb/bigchaindb.git
```
sohkai: I didn't want to change this, but if we'll change this back later then it's fine.

tests/test_connection.py
```diff
@@ -18,3 +28,17 @@ def test_response_content_type_handling(self, content_type, json, data):
...
+    def test_request_with_headers(self, headers):
+        from bigchaindb_driver.connection import Connection
+        url = 'http://dummy'
+        connection = Connection(node_url=url, headers=headers)
+        with RequestsMock() as requests_mock:
+            requests_mock.add('GET', url, adding_headers=headers)
+            response = connection.request('GET')
+        assert response.status_code == 200
+        del response.headers['Content-type']
+        assert response.headers == headers
```
sohkai: Rather than doing a full compare, maybe it'd be safer to just check that the custom headers are in `response.headers`?

tests/test_driver.py
```diff
@@ -3,26 +3,34 @@
...
+    def test_driver_init(self, nodes, headers):
+        from bigchaindb_driver.driver import BigchainDB, DEFAULT_NODE
+        driver = BigchainDB(*nodes, headers=headers)
+        nodes = (DEFAULT_NODE,) if not nodes else nodes
+        headers = {} if not headers else headers
+        assert driver.nodes == nodes
+        assert driver.transport.nodes == nodes
+        for conn in driver.transport.pool.connections:
+            expected_headers = default_headers()
+            expected_headers.update(headers)
```
sohkai: Can these two lines be moved outside the loop?

bigchaindb_driver/driver.py
sbellem: ah, ok

compose/server/Dockerfile
sbellem: yes, I do not really know a better way

tests/test_connection.py
sbellem: Ah, yes. The reason is to make sure that the headers added by the `requests` lib are not lost, hence the full check.

tests/test_driver.py
sbellem: Indeed!

bigchaindb_driver/connection.py
sbellem: Nope. They are added to the ones provided by the `Session` object.  See https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L403

PR #227, opened 2017-01-26
Fix `prepare_create_transaction` leaving `data` in asset

sohkai:
`prepare_create_transaction` should strip out the `data` from the provided `asset` to pass into `Transaction.create()`.

### DIFF
--- a/tests/test_offchain.py
+++ b/tests/test_offchain.py
@@ -32,6 +32,7 @@ def test_prepare_create_transaction_default(alice_pubkey):
     assert 'id' in create_transaction
     assert 'version' in create_transaction
     assert 'asset' in create_transaction
+    assert create_transaction['asset'] == {'data': None}
     assert 'outputs' in create_transaction
     assert 'inputs' in create_transaction
     assert 'metadata' in create_transaction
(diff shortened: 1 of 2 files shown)


tests/test_offchain.py
```diff
@@ -29,9 +29,11 @@ def test_prepare_transaction_raises():
 def test_prepare_create_transaction_default(alice_pubkey):
     from bigchaindb_driver.offchain import prepare_create_transaction
     create_transaction = prepare_create_transaction(signers=alice_pubkey)
+    del create_transaction['asset']['id']
```
sohkai: I've had to add these to make tests pass until bigchaindb/bigchaindb#1101 is fixed.

tests/test_offchain.py
```diff
@@ -40,7 +42,7 @@ def test_prepare_create_transaction_default(alice_pubkey):
 
 
 @mark.parametrize('asset', (
-    None, {}, {'data': {'msg': 'Hello BigchainDB!'}},
+    None, {'data': None}, {'data': {'msg': 'Hello BigchainDB!'}},
```
sohkai: Note that any `asset` dict being passed into `prepare_transaction()` needs to be either of the following formats:

* `CREATE`:
```json
{
    "data": ...
}
```

* `TRANSFER`:
```json
{
    "id": ...
}
```

PR #231, opened 2017-01-26
Update unspents endpoint to latest changes

sbellem:
closes #213

PR #232, opened 2017-01-27
Additional docstrings changes for 0.9

sohkai:
Mostly updates `BigchainDB` class's docstrings to match current behaviour.

Contains some other small typo fixes / adjustments.

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -32,7 +32,7 @@ class BigchainDB:
             headers (dict): Optional headers that will be passed with
                 each request. To pass headers only on a per-request
                 basis, you can pass the headers to the method of choice
-                (e.g. :meth:`~.TransactionsEndpoint.send`).
+                (e.g. :meth:`.TransactionsEndpoint.send`).
 
         """
         self._nodes = nodes if nodes else (DEFAULT_NODE,)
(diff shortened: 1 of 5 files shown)


bigchaindb_driver/utils.py
```diff
@@ -1,10 +1,12 @@
 """Set of utilities to support various functionalities of the driver.
 
+.. danger:: This module is meant for specific internal usage only and its
+    behaviour is subject to change.
+
```
sohkai: Moved the danger notice to be module-level since I don't think anyone should be using any of this directly.

docs/libref.rst
```diff
@@ -16,8 +16,6 @@ Library Reference
 .. autoclass:: TransactionsEndpoint
     :members:
 
-    .. automethod:: __init__
-
```
sohkai: I don't think there's any point in including the `NamespacedDriver.__init__` in here.

bigchaindb_driver/driver.py
```diff
@@ -244,10 +239,10 @@ def status(self, txid, headers=None):
 
 
 class UnspentsEndpoint(NamespacedDriver):
-    """Endpoint for unspents.
+    """Exposes functionality of the ``'/unspents'`` endpoint.
 
     Attributes:
-        path (str): The path of the endpoint: ``'/unspents'``
+        path (str): The path of the endpoint.
```
sohkai: Note that these changes may slightly conflict with #231.

bigchaindb_driver/utils.py
sbellem: We can remove this notice all together, as discussed offline.

Issue #105, opened 2016-10-13
Fix intersphinx inventory 'objects.inv' not fetchable


sbellem:
``` bash
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://docs.bigchaindb.com/projects/server/en/latest/objects.inv...
WARNING: intersphinx inventory 'https://docs.bigchaindb.com/projects/server/en/latest/objects.inv' not fetchable due to <class 'urllib.error.HTTPError'>: HTTP Error 403: Forbidden
```
### Notes

This was working last week. Not sure why it does not now. One thing that is new is that [docs.bigchaindb.com](https://docs.bigchaindb.com) is now available under `https`. Could that be related? Needs investigations.
### Suggested troubleshooting approach

In order to troubleshoot this issue, one may simply use the `fetch_inventory` method from `intersphinx`:

``` python
>>> import warnings
>>> from sphinx.ext.intersphinx import fetch_inventory
>>> uri = 'https://docs.python.org/3/'
>>> fetch_inventory(warnings, uri, uri + 'objects.inv')
  # ..., ..., ..., ..., ..., ...,
  'yield_atom': ('Python',
   '3.5',
   'https://docs.python.org/3/reference/expressions.html#grammar-token-yield_atom',
   '-'),
  'yield_expression': ('Python',
   '3.5',
   'https://docs.python.org/3/reference/expressions.html#grammar-token-yield_expression',
   '-'),
  'yield_stmt': ('Python',
   '3.5',
   'https://docs.python.org/3/reference/simple_stmts.html#grammar-token-yield_stmt',
   '-')}}
```

As shown above, it should work for Python. If you now try with BigchainDB:

``` python
>>> uri = 'https://docs.bigchaindb.com/projects/server/en/latest/'
>>> fetch_inventory(warnings, uri, uri + 'objects.inv')
/usr/local/lib/python3.5/site-packages/sphinx/ext/intersphinx.py:238: UserWarning: intersphinx inventory 'https://docs.bigchaindb.com/projects/server/en/latest/objects.inv' not fetchable due to <class 'urllib.error.HTTPError'>: HTTP Error 403: Forbidden
  '%s: %s' % (inv, err.__class__, err))
```

Other variations of `uri` have been tried and all failed as well:

``` python
In [37]: uri = 'https://docs.bigchaindb.com/projects/server/en/'

[...]

PR #234, opened 2017-01-27
Update basic usage docs for 0.9

sohkai:
In general I've tried to make the language consistent with what the server uses (e.g. "outputs" instead of "conditions").

Note that only https://github.com/bigchaindb/bigchaindb-driver/pull/234/commits/9343ebea186555520e5f5f74334542abd3547faf and https://github.com/bigchaindb/bigchaindb-driver/pull/234/commits/cb83a9d4c9e17e2d37d8eb4c5a861535da2a2b95 contain non-whitespace changes.

-----------

### Preview:

basic usage examples

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -25,7 +25,7 @@ Connecting to a BigchainDB node is done via the
     bdb = BigchainDB('<api_endpoint>')
 
 where ``<api_endpoint>`` is the root URL of the BigchainDB server API you wish
-to connect to. 
+to connect to.
 
 For the simplest case in which a BigchainDB node would be running locally, (and
 the ``BIGCHAINDB_SERVER_BIND`` setting wouldn't have been changed), you would
@@ -48,7 +48,7 @@ Alternatively, you may connect to the containerized BigchainDB node from
 "outside", in which case you need to know the port binding:
 
 .. code-block:: bash
-    
+
     $ docker-compose port bdb-server 9984
     0.0.0.0:32780
 
@@ -71,7 +71,7 @@ As an example, let's consider the creation and transfer of a digital asset that
 represents a bicycle:
 
 .. ipython::
-    
+
     In [0]: bicycle = {
        ...:     'data': {
        ...:         'bicycle': {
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -12,11 +12,12 @@ Basic Usage Examples
 
 
 The BigchainDB driver's main purpose is to connect to one or more BigchainDB
-server nodes, in order to perform supported API calls documented under
-:doc:`drivers-clients/http-client-server-api`.
+server nodes, in order to perform supported API calls (see the server's
+`HTTP API docs <https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/http-client-server-api.html>`_
```
sohkai: Not sure if there's a better way to link to this now?

sbellem: I think we need to handle https://github.com/bigchaindb/bigchaindb-driver/issues/105

docs/usage.rst
```diff
@@ -199,31 +199,44 @@ or simply use ``fulfilled_creation_tx``:
...
-    In [0]: condition = creation_tx['conditions'][cid]
+    In [0]: transfer_asset = {
+       ...:     'id': asset_id,
+       ...: }
+
+Let's now prepare the transfer transaction:
+
+.. ipython::
+
+    In [0]: fulfilled_output_id = 0
```
sbellem: How about calling it `index` instead of `id` to make it clearer, since it is really the `index` of the element of a list.

sbellem: i.e.: `fulfilled_output_index`

docs/usage.rst
```diff
@@ -199,31 +199,44 @@ or simply use ``fulfilled_creation_tx``:
...
+
+    In [0]: fulfilled_output = creation_tx['outputs'][fulfilled_output_id]
```
sbellem: `output` and `output_index` would be simpler and less confusing as the output has not been fulfilled yet -- we can assume (or hope) that the reader will understand with the help of the context of the example

PR #235, opened 2017-01-30
Update advanced usage docs for 0.9

sohkai:
Similar to #234.

Note that only https://github.com/bigchaindb/bigchaindb-driver/commit/0ce630daf23bfddc453ee04d31655fc19d6a5d38, https://github.com/bigchaindb/bigchaindb-driver/commit/1675ae588051a0296c6568d27facbe8fd8b854ee, and https://github.com/bigchaindb/bigchaindb-driver/commit/88df45c7859665dbbe58ddf1615eafe1d0ffde5b contain non-whitespace changes.

**Note**: I've avoided changing the sections with todo (e.g. hashlocks, thresholds, timelocks, etc) aside from modifying any direct access to a transaction that had become obsolete (see https://github.com/bigchaindb/bigchaindb-driver/commit/88df45c7859665dbbe58ddf1615eafe1d0ffde5b).

--------

### Preview:

image

PR #236, opened 2017-01-30
Update handcrafting usage docs for 0.9

sohkai:
Similar to #234 and #235.

No preview because it's too big 😢

### DIFF
--- a/docs/handcraft.rst
+++ b/docs/handcraft.rst
@@ -95,8 +95,8 @@ From the point of view of Python, a transaction is simply a dictionary:
         'version': 1
     }
 
-Because a transaction must be signed before being sent, the ``id``, must be
-provided by the client.
+Because a transaction must be signed before being sent, the ``id`` and
+``fulfillment`` must be provided by the client.
 
 .. important:: **Implications of Signed Payloads**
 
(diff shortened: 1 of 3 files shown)


docs/handcraft.rst
```diff
@@ -897,15 +912,15 @@ The Fulfilled Transaction
 
     In [0]: ed25519.sign(message.encode(), sk)
 
-    In [0]: fulfillment = ed25519.serialize_uri()
+    In [0]: fulfillment_uri = ed25519.serialize_uri()
```
sbellem: As a side note: the reason for which the variable was named `fulfillment` and not `fulfillment_uri` is because the key for which it holds the value is named `fulfillment` and not `fulfillment_uri`.

docs/handcraft.rst
```diff
@@ -1749,34 +1750,32 @@ Handcrafting the ``'TRANSFER'`` transaction
...
     handcrafted_car_transfer_tx = {
+        'operation': 'TRANSFER',
         'asset': {'id': car_asset['id']},
         'metadata': None,
-        'operation': 'TRANSFER',
-        'conditions': (condition,),
-        'fulfillments': (fulfillment,),
+        'outputs': (output,),
+        'inputs': (input_,),
         'version': 1,
```
sbellem: version needs to be set to `0.9`

As an aside I think simply assigning the version in the `dict` is okay, that is, without the need to set a variable beforehand.

PR #238, opened 2017-01-31
Small things

sbellem:


bigchaindb_driver/driver.py
```diff
@@ -27,7 +27,8 @@ def __init__(self, *nodes, transport_class=Transport, headers=None):
             headers (dict): Optional headers that will be passed with
                 each request. To pass headers only on a per-request
                 basis, you can pass the headers to the method of choice
-                (e.g. :meth:`~.TransactionsEndpoint.send`).
+                (e.g. :meth:`BigchainDB().transactions.send()
```
sohkai: I find the `BigchainDB()` a little bit confusing but having just `.transactions.send()` is only marginally better.

sbellem: Oh! @sohkai just realized after merging I forgot to make the small proposed change. Will do in a a separate PR.

PR #239, opened 2017-01-31
Transaction model's version must be a string

sohkai:
Oops, missed this in #236.


sbellem:
#240 should solve the error on travis

PR #251, opened 2017-02-02
New docs page w/ examples of how to connect

### LINKED ISSUE #152: New docs page about how to connect to various API endpoint server options
I noticed that the **Basic Usage Examples** page and the **Advanced Usage [Examples]** page both start out with some information about how to make a connection to some BigchainDB API endpoint server running somewhere.

I wonder if it would make sense to put that on a new separate page between **Quickstart / Installation** and **Basic Usage Examples**. Maybe it could be titled **Step 2: Connect**. It could have several sections, one for each case, starting with the easiest and most common. The **Basic Usage Examples** page and the **Advanced Usage [Examples]** page could start by saying they must be connected to a BigchainDB API endpoint server running somewhere, with a link to the new page if they need to do that.

ttmc:
This resolves #152 

I also changed the way that H1, H2 and H3 headings are created in `advanced-usage.rst`, to be consistent with the other pages.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -4,66 +4,9 @@
 Basic Usage Examples
 ====================
 
-.. note::
-
-   You must :doc:`install the bigchaindb_driver Python package <quickstart>` first.
-
-   You should use Python 3 for these examples.
-
-
-The BigchainDB driver's main purpose is to connect to one or more BigchainDB
-server nodes, in order to perform supported API calls (see the server's
-`HTTP API docs <https://docs.bigchaindb.com/projects/server/en/latest/drivers-clients/http-client-server-api.html>`_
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -4,66 +4,9 @@
...
-
-    bdb = BigchainDB('http://0.0.0.0:32780/api/v1')
-
-For the sake of this example, we'll assume:
-
-.. ipython::
-
-    In [0]: from bigchaindb_driver import BigchainDB
-
-    In [0]: bdb = BigchainDB('http://bdb-server:9984/api/v1')
```
sohkai: A small thing here -- I think we still need to have this last iPython prompt as the rest of the document relies on having access to a `bdb` object.

Issue #250, opened 2017-02-02
For basic usage examples: Have recap section to allow easy copy and paste


TimDaub:
Like shown here: https://docs.bigchaindb.com/projects/py-driver/en/latest/handcraft.html#in-a-nutshell

have a recap section for basic use case examples.

sbellem:
```rst
In a nutshell
"""""""""""""

.. code-block:: python
    # ...
```

Issue #209, opened 2016-12-18
Problem with while loop in Basic Usage Examples when done by script


ttmc:
This problem was reported by Thomas from Microsoft. I reproduced it on my machine.

* bigchaindb version: 0.8.0
* bigchaindb-driver version: 0.1.3
* Python version: 3.5.2
* Operating System: Ubuntu 14.04

### Setup

1. Install the versions as above.
2. Run RethinkDB and BigchainDB locally.
3. Run the following Python script (based on the Python Driver Basic Usage Examples):
```python
from bigchaindb_driver import BigchainDB
from bigchaindb_driver.crypto import generate_keypair

bdb = BigchainDB('http://localhost:9984/api/v1')

alice, bob = generate_keypair(), generate_keypair()
bicycle = {'data': {'bicycle': {'serial_number': 'abcd1234','manufacturer': 'bkfab'}}}
metadata = {'planet': 'earth'}

prepared_creation_tx = bdb.transactions.prepare(operation='CREATE', owners_before=alice.verifying_key, asset=bicycle, metadata=metadata)

fulfilled_creation_tx = bdb.transactions.fulfill(prepared_creation_tx, private_keys=alice.signing_key)

print(fulfilled_creation_tx)

sent_creation_tx = bdb.transactions.send(fulfilled_creation_tx)

txid = fulfilled_creation_tx['id']

print(txid)

print(sent_creation_tx == fulfilled_creation_tx)

trials = 0
while bdb.transactions.status(txid).get('status') != 'valid' and trials < 100:
    trials += 1

print(trials)
print(bdb.transactions.status(txid))
```

### What Happened

Here's my terminal output:
```bash
(p35bd) troy@ascribe3-XPS-8700:~/repos/bigchaindb-driver$ python ~/Desktop/test2.py 
[...]

PR #252, opened 2017-02-02
Include code example

krish7919:
- Added a working/complete code example on asset creation and transfer in docs

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -348,14 +348,18 @@ Recap: Asset Creation & Transfer
 	txid = fulfilled_creation_tx['id']
 	
 	trials = 0
-	while bdb.transactions.status(txid).get('status') != 'valid' and trials < 60:
-		trials += 1
-		sleep(1)
+	while trials < 60:
+		try:
+			if bdb.transactions.status(txid).get('status') == 'valid':
+				print('Tx valid in:', trials, 'secs')
+				break
+		except bigchaindb_driver.exceptions.NotFoundError:
+			trials += 1
+			sleep(1)
 	
 	if trials == 60:
 		print('Tx is still being processed... Bye!')
 		exit(0)
-	print('Tx valid in:', trials, 'secs')
 	
 	asset_id = txid
 	
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -192,6 +192,7 @@ Here's some code that keeps checking the status of the transaction until it is v
 
 .. _bicycle-transfer:
 
+
 Asset Transfer
 --------------
 Imagine some time goes by, during which Alice is happy with her bicycle, and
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -302,6 +303,99 @@ Alice is the former owner:
...
+		prepared_creation_tx,
+		private_keys=alice.private_key
+	)
+	
+	sent_creation_tx = bdb.transactions.send(fulfilled_creation_tx)
+	
+	txid = fulfilled_creation_tx['id']
+	
+	trials = 0
+	while bdb.transactions.status(txid).get('status') != 'valid' and trials < 60:
```
ttmc: This line will probably fail at first. The latest version of this page (`usage.rst`) puts a try/except inside the loop. (It also uses 100 instead of 60, but that doesn't matter.) Are you working from the latest `master` branch?

docs/usage.rst
```diff
@@ -302,6 +303,99 @@ Alice is the former owner:
...
+		trials += 1
+		sleep(1)
+	
+	if trials == 60:
+		print('Tx is still being processed... Bye!')
```
ttmc: Another possibility is that the transaction was invalid and got deleted from the system.

docs/usage.rst
```diff
@@ -302,6 +303,99 @@ Alice is the former owner:
...
+		exit(0)
+	print('Tx valid in:', trials, 'secs')
```
ttmc: This line will always print `Tx valid in: something secs` even if the transaction was invalid. I'd just remove this line.

docs/usage.rst
krish7919: It will fail and will be retried for 60 secs, before the script gives up and exits.

krish7919: I tested the script with the latest bdb server locally and it works. Sometimes it blocks for upto 4 secs, so 60 secs seems like a nice value during demos and trials, etc.

docs/usage.rst
krish7919: If the transaction was invalid, the script will exit, and the line won't be printed.

docs/usage.rst
krish7919: Well, in that case, I can change the message to say 'Tx has not been validated for 60 secs... Bye!'.
It's a message for exiting the script in case the tx we passed to the server isn't valid in 60 secs.

docs/usage.rst
ttmc: I added the try/except inside the loop to fix a problem that was reported by a user (in issue #209). Basically, evaluating the expression `bdb.transactions.status(txid).get('status')` can raise an exception before it can be compared with `'valid'`. I resolved that issue in pull request #244.

docs/usage.rst
ttmc: Such a message would be misleading because sometimes the transaction _was_ validated (i.e. checked for validity) and found to be invalid.

docs/usage.rst
ttmc: Ah, ok.

docs/usage.rst
krish7919: Handled.

docs/usage.rst
```diff
@@ -192,6 +192,7 @@ Here's some code that keeps checking the status of the transaction until it is v
 
 .. _bicycle-transfer:
 
+
```
sohkai: Let's avoid adding the whitespace here (note the `ref` directive above).

docs/usage.rst
```diff
@@ -302,6 +303,103 @@ Alice is the former owner:
         transfer_asset_id = transfer_tx['asset']['id']
 
 
+
```
sohkai: And I think this extra line could be removed too.

docs/usage.rst
```diff
@@ -302,6 +303,103 @@ Alice is the former owner:
...
+.. code-block:: python
+
+	from bigchaindb_driver import BigchainDB
+	from bigchaindb_driver.crypto import generate_keypair
+	from time import sleep
+	from sys import exit
+	
+	alice, bob = generate_keypair(), generate_keypair()
+	
+	bdb = BigchainDB('http://localhost:59984/api/v1')
```
sohkai: Probably just `BigchainDB()` is fine for the example.

docs/usage.rst
```diff
@@ -302,6 +303,103 @@ Alice is the former owner:
...
+	)
+	
+	sent_transfer_tx = bdb.transactions.send(fulfilled_transfer_tx)
+	
+	print("Is Bob the owner?",
+		sent_transfer_tx['outputs'][0]['public_keys'][0] == bob.public_key)
+	
+	print("Was Alice the previous owner?",
+		fulfilled_transfer_tx['inputs'][0]['owners_before'][0] == alice.public_key)
+
```
sohkai: I think it might be nice to add a few comments in this block, to denote separate parts that people might be interested in.

For example, we could denote that we're going to make a `CREATE` transaction, then note that we'll be sending it (`bdb.transactions.send`), and then check its status (`bdb.transactions.status`), etc.

ttmc: As I understand it, this code is an almost-repeat of above-explained code, in one convenient block.

docs/usage.rst
krish7919: We want a working example here. Mostly for Dimi's demos. :)

docs/usage.rst
krish7919: @ttmc : This is a working recap of the previous code snippets showing both asset creation and transfer. Easier for people to follow entire code flow after snippets.

krish7919:
Solves #250

docs/usage.rst
sohkai: It might be one convenient block, but it's not meant as a copy pasta script. I imagine people would want to scan it, find the sections they care about, and then copy those.

PR #253, opened 2017-02-02
Get info of a node and its api root

sbellem:
Add support for:

* `GET /`
* `GET /api/{version}`

via the `BigchainDB` class, which now only requires the scheme, hostname and port (e.g.: http://localhost:9984')

- [x] Document new methods (`node_info`, `api_info`)
- [x] Update connection docs

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -64,6 +64,49 @@ class BigchainDB:
         """
         return self._outputs
 
+    def info(self, headers=None):
+        """Retrieves information of the node being connected to via the
+        root endpoint ``'/'``.
+
+        Args:
+            headers (dict): Optional headers to pass to the request.
+
+        Returns:
+            dict: Details of the node that this instance is connected
+            to. Some information that may be interesting:
+
+                * the server version,
+                * the public key of the node, and
+                * its key ring (list of public keys of the nodes this
+                  node is connected to).
+
+        Note:
+            Currently limited to one node, and will be expanded to
+            return information for each node that this instance is
+            connected to.
+
+        """
+        return self.transport.forward_request(
+            method='GET', path='/', headers=headers)
+
+    def api_info(self, headers=None):
+        """Retrieves information provided by the API root endpoint
+        ``'/api/v1'``.
+
+        Args:
+            headers (dict): Optional headers to pass to the request.
+
+        Returns:
+            dict: Details of the HTTP API provided by the BigchainDB
+            server.
+
+        """
+        return self.transport.forward_request(
+            method='GET',
+            path=self.api_prefix,
+            headers=headers,
+        )
+
 
 class NamespacedDriver:
     """Base class for creating endpoints (namespaced objects) that can be added
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -2,7 +2,7 @@ from .transport import Transport
 from .offchain import prepare_transaction, fulfill_transaction
 
 
-DEFAULT_NODE = 'http://localhost:9984/api/v1'
+DEFAULT_NODE = 'http://localhost:9984'
 
 
 class BigchainDB:
@@ -21,7 +21,7 @@ class BigchainDB:
         Args:
             *nodes (str): BigchainDB nodes to connect to. Currently, the full
                 URL must be given. In the absence of any node, the default
-                (``'http://localhost:9984/api/v1'``) will be used.
+                (``'http://localhost:9984'``) will be used.
             transport_class: Optional transport class to use.
                 Defaults to :class:`~bigchaindb_driver.transport.Transport`.
             headers (dict): Optional headers that will be passed with
@@ -35,6 +35,7 @@ class BigchainDB:
         self._transport = transport_class(*self._nodes, headers=headers)
         self._transactions = TransactionsEndpoint(self)
         self._outputs = OutputsEndpoint(self)
+        self.api_prefix = '/api/v1'
 
     @property
     def nodes(self):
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python: 3.5
 
 env:
+  - BDB_PUKEY=GW1nrdZm4mbVC8ePeiGWz6DqHexqewqy5teURVHi3RG4
```
krish7919: Is that BDB_PUBKEY?

bigchaindb_driver/driver.py
```diff
@@ -35,6 +36,8 @@ def __init__(self, *nodes, transport_class=Transport, headers=None):
         self._transport = transport_class(*self._nodes, headers=headers)
         self._transactions = TransactionsEndpoint(self)
         self._outputs = OutputsEndpoint(self)
+        self.api_version = api_version
+        self.api_prefix = '/api/{}'.format(api_version)
```
sbellem: let's just assume `'v'` will _always_ be, so we can pass a simple integer for the api version

bigchaindb_driver/driver.py
```diff
@@ -63,11 +66,24 @@ def outputs(self):
...
 
+    def node_info(self, headers=None):
+        return self.transport.forward_request(
+            method='GET', path='/', headers=None)
+
+    def api_info(self, headers=None):
+        return self.transport.forward_request(
+            method='GET',
+            path=self.api_prefix,
+            headers=None,
```
sohkai: The headers in the `forward_request()` for `node_info` and `api_info` should be `headers` instead of `None`?

tests/conftest.py
```diff
@@ -171,8 +171,13 @@ def bdb_port():
 
 
 @fixture
+def bdb_node_pubkey():
+    return environ['BIGCHAINDB_KEYPAIR_PUBLIC']
```
sohkai: Mmm, if someone runs this test locally, this value won't be correct. Let's just hardcode it for now.

Alternatively, we don't bother testing for the exact keypair in `test_node_info`.

sbellem:
> Also, the documentation should be changed to reflect that the expected node format has changed.

this is mentioned in https://github.com/bigchaindb/bigchaindb-driver/pull/253#issue-204922782

ttmc:
Looking okay so far...

bigchaindb_driver/driver.py
```diff
@@ -65,10 +65,42 @@ def outputs(self):
         return self._outputs
 
     def node_info(self, headers=None):
+        """Retrivues information of the node being connected to via the
```
sohkai: > **Retrieves**

bigchaindb_driver/driver.py
```diff
@@ -65,10 +65,42 @@ def outputs(self):
...
+        root endpoint ``'/'``.
+
+        Args:
+            headers (dict): Optional headers to pass to the request.
+
+        Returns:
+            dict: Details of the node that this instance is connected
+            to. Some information that may be interesting comprises:
```
sohkai: > that may be interesting: ~~comprises~~

docs/connect.rst
```diff
@@ -18,9 +18,8 @@ class so we can use it to connect:
 
     from bigchaindb_driver import BigchainDB
 
-Finally, to make the connection, you need to know
-the BigchainDB "API Root Endpoint"
-of a BigchainDB node or cluster.
+Finally, to make the connection, you need to know the URL address of the
```
ttmc: The BigchainDB HTTP API docs call this (e.g. `http://example.com:9984`) the "BigchainDB Root URL", e.g. see https://docs.bigchaindb.com/projects/server/en/master/drivers-clients/http-client-server-api.html#bigchaindb-root-url

docs/advanced-usage.rst
```diff
@@ -49,7 +49,7 @@ an instance:
 This instantiates an object ``bdb`` of class
 :class:`~bigchaindb_driver.BigchainDB`. When instantiating a
 :class:`~bigchaindb_driver.BigchainDB` object without arguments (as above), it
-uses the default BigchainDB API Root Endpoint ``http://localhost:9984/api/v1``.
+uses the default BigchainDB API Root Endpoint ``http://localhost:9984``.
```
ttmc: The BigchainDB Server docs call this the "BigchainDB Root URL"

docs/advanced-usage.rst
```diff
@@ -49,7 +49,7 @@ an instance:
 This instantiates an object ``bdb`` of class
 :class:`~bigchaindb_driver.BigchainDB`. When instantiating a
 :class:`~bigchaindb_driver.BigchainDB` object without arguments (as above), it
-uses the default BigchainDB API Root Endpoint ``http://localhost:9984/api/v1``.
+uses the default BigchainDB Root URL Endpoint ``http://localhost:9984``.
```
ttmc: I'd prefer "BigchainDB Root URL" (with no "Endpoint" on the end) but I guess this will do.

bigchaindb_driver/driver.py
```diff
@@ -63,11 +64,56 @@ def outputs(self):
...
+
+                * the server version,
+                * the public key of the node, and
+                * its key ring (list of public keys of the nodes this
+                  node is connected to).
+
+        Note:
+            Currently limited to one node, and will be expanded to
+            return information for each node that the instance this
+            connected to.
```
krish7919: AFAIK, we do not have a way to connect to all the instances in the federation. Is there one?
I also thought that to connect to all the instances, the driver would need to be provided with an explicit list of instances.

ttmc:
A test failed in TravisCI:

screenshot from 2017-02-06 17-34-11

PR #254, opened 2017-02-03
Add back `bdb` instance into basic usage examples

sohkai:
I do think we still need this bit... see https://github.com/bigchaindb/bigchaindb-driver/pull/251/files#r99296405.

Fixes some extraneous whitespace and tabs as well. Only commit that has non-cosmetic changes is 534961e.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -129,7 +129,7 @@ Here's some code that keeps checking the status of the transaction until it is v
     ...             break
     ...     except bigchaindb_driver.exceptions.NotFoundError:
     ...         trials += 1
-    
+
     >>> bdb.transactions.status(txid)
     {'status': 'valid'}
 
@@ -256,11 +256,11 @@ Recap: Asset Creation & Transfer
 	from bigchaindb_driver.crypto import generate_keypair
 	from time import sleep
 	from sys import exit
-	
+
 	alice, bob = generate_keypair(), generate_keypair()
-	
+
 	bdb = BigchainDB('http://localhost:59984/api/v1')
-	
+
 	bicycle_asset = {
 		'data': {
 			'bicycle': {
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -8,6 +8,14 @@ First, make sure you're using Python 3,
 you've :doc:`installed the bigchaindb_driver Python package <quickstart>`,
 and you've :doc:`connected to a BigchainDB node or cluster <connect>`.
 
+For the sake of these examples, we'll assume:
+
+.. ipython::
+
+    In [0]: from bigchaindb_driver import BigchainDB
+
+    In [0]: bdb = BigchainDB('http://bdb-server:9984/api/v1')
```
ttmc: The whole point of the new `connect.rst` page was to allow for all the diverse possible URL strings that could go inside the first argument of `BigchainDB()`. Why do we need to assume that it's `'http://bdb-server:9984/api/v1'`? I already told them to assume that they're connected to a BigchainDB node or cluster, with a link to the `connect.rst` page.

Maybe the solution is to edit the `connect.rst` page to tell them to use IPython and to put IPython code blocks everywhere on that page?

sohkai: Mmm, not sure how the IPython directives work, but I imagine their scope is limited to their current page. Maybe @sbellem knows, but likely, we'll need to put this bit *somewhere* on this page.

sohkai: > Why do we need to assume that it's 'http://bdb-server:9984/api/v1'?

No particular reason, aside from that's what it was before, and that's what another example uses later in the document. We could also use the default node, via `BigchainDB()`, so it's less confusing?

ttmc: Okay, so maybe comment that `'http://bdb-server:9984/api/v1'` is just one of many possible URLs and they should use _their URL_, as shown on the `connect.rst` page?

PR #284, opened 2017-05-04
Upgrade server version

sbellem:


setup.py
```diff
@@ -11,7 +11,7 @@
 
 install_requires = [
     'requests>=2.11.0',
-    'bigchaindb~=0.9.1',
+    'bigchaindb>=0.10',
```
ttmc: This might cause problems for someone who wants to use version 0.2.2 of the Python driver (for some reason) once the latest `bigchaindb` (server) package is 1.x, since 1.x will have breaking changes in the transaction model, but those won't be in the docs for 0.2.2. My inclination would be to use:
```
    'bigchaindb~=0.10.1',
```

PR #286, opened 2017-05-11
Docker files updates

sbellem:
This PR updates the docker-compose file so that:

1. MongoDB can be used as a backend when running BigchainDB server.
2. The websocket events API can be used.

Note for @ttmc this will most likely require minor changes to docs. Will add you as a reviewer once the docs have been updated.

- [x] update docs

### DIFF
--- a/compose/driver/Dockerfile
+++ b/compose/driver/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.1
+FROM python:3.6.2rc1
 
 RUN apt-get update && apt-get install -y vim
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -73,19 +73,19 @@ development.
    Now you can make your changes locally.
 
 4. When you're done making changes, check that your changes pass flake8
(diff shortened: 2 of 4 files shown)


compose/driver/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6.1
+FROM python:3.6.2rc1
```
krish7919: Should we have an `rc` as base image?

libscott:
I like the simpler docker-compose file, with `db` and `bdb-server`. I wonder if `bdb` isn't a little too general, though, since it's actually the driver that it's running.

compose/driver/Dockerfile
sbellem: rc2 is being used here because that is what travis uses

remember that this is for development purposes

sbellem:
> I wonder if bdb isn't a little too general, though, since it's actually the driver that it's running.

This is from a client point of view. We have been using `bdb` in the client's docs to refer to an instance of the client for example.

PR #294, opened 2017-06-16
Support HTTP API v1.0

### LINKED ISSUE #291: Update the driver to support the breaking changes introduced in the HTTP API
Update the driver to support the breaking changes introduced in the HTTP API for the next release of bigchaindb.

The changes are documented here https://github.com/bigchaindb/bigchaindb/blob/babe5311a46684ef7679ec843b1f600a7bceed38/docs/upgrade-guides/v0.10--%3Ev1.0.md

The proposals are documented here https://github.com/bigchaindb/bigchaindb/blob/99499b1f8783719a082813912ac9a0d363ae278f/bdb-ip.md

And the issues regarding the changes are under this milestone https://github.com/bigchaindb/bigchaindb/milestone/11?closed=1

r-marques:
resolves #291 

This updates the driver to support the breaking changes introduced in the HTTP API to be released with version 1.0 of BigchainDB.

This was tested against BigchainDB master branch with all the breaking changes https://github.com/bigchaindb/bigchaindb/commit/571ee83731d9ed5a80aae65ab612bae52a354f98

The BigchainDB dependency of the bigchaindb-driver is also using the same code.

- Update info endpoints
- Amount is now a string
- Output endpoint:
  - `unspent` -> `spent`
  - ability to only return spent outputs
- Rename of `txid` and `tx_id` to `transaction_id`
  - Data models
  - http arguments
- Updated the documentation
- Rename all occurrences of `txid` to `transaction_id`, again!

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -341,7 +341,8 @@ class TransactionsEndpoint(NamespacedDriver):
         # HTTP API more thoroughly.
         path = self.api_prefix + '/statuses'
         return self.transport.forward_request(
-            method='GET', path=path, params={'tx_id': txid}, headers=headers)
+            method='GET', path=path, params={'transaction_id': txid},
+            headers=headers)
 
 
 class OutputsEndpoint(NamespacedDriver):
@@ -353,15 +354,16 @@ class OutputsEndpoint(NamespacedDriver):
     """
     PATH = '/outputs/'
 
-    def get(self, public_key, unspent=False, headers=None):
+    def get(self, public_key, spent=None, headers=None):
         """
 
         Args:
             public_key (str): Public key for which unfulfilled
                 conditions are sought.
-            unspent (bool): Whether to get the unconsumed outputs
-                only. Defaults to ``False``, meaning that both consumed,
-                and unconsumed outputs will be returned.
+            spent (bool): Indicate if the result set should include only spent
+                or only unspent outputs. If not specified (``None``) the
+                result includes all the outputs (both spent and unspent)
+                associated with the public_key
             headers (dict): Optional headers to pass to the request.
 
         Returns:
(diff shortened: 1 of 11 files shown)


bigchaindb_driver/driver.py
```diff
@@ -353,15 +354,16 @@ class OutputsEndpoint(NamespacedDriver):
...
         Args:
             public_key (str): Public key for which unfulfilled
                 conditions are sought.
-            unspent (bool): Whether to get the unconsumed outputs
-                only. Defaults to ``False``, meaning that both consumed,
-                and unconsumed outputs will be returned.
+            spent (bool): Indicate if the result set should include only spent
+                or only unspent outputs. If not specified (``None``) the
+                result includes all the outputs (both spent and unspent)
+                associated with the public_key
```
sbellem: Very small details: 

1. public key or ``public_key``
2. end of sentence with a period "."

Sorry for being so picky :)

tests/conftest.py
```diff
@@ -29,7 +29,7 @@ def wrapper(*args, **kwargs):
 def make_ed25519_condition(public_key, *, amount=1):
     ed25519 = Ed25519Fulfillment(public_key=public_key)
     return {
-        'amount': amount,
+        'amount': str(amount),
```
TimDaub: I know, this is just a conftest function, however:

I took a different approach regarding passing the amount. In Javascript, every amount now being passed to some function needs to be of type string, to avoid integer overflows on the user-side:


```python
# python max integer (within the range of asset.amount's max value)
>>> sys.maxint
9223372036854775807
```

```javascript
// js max integer (not within the range of asset.amount's max value)
Number.MAX_SAFE_INTEGER // 9007199254740991
```

Since Python's max integer is well within asset.amount's max value, we can still pass integers to user-facing functions, where as in Javascript we should always fall back to strings or even an extra library (e.g. big.js).


Just wanted to share that information.

r-marques:
As far as I can see the occurrences in `advanced-usage.rst` are correct

https://github.com/bigchaindb/bigchaindb-driver/blob/master/bigchaindb_driver/offchain.py#L316

PR #297, opened 2017-06-22
Cryptoconditions 02 support

r-marques:
This pr adds support for cryptoconditions draft 02 spec.

- Updated fixtures
- Fixed tests
- Updated handcrafting transactions


r-marques:
Before releasing we also need to update the compatibility matrix in the documentation

PR #309, opened 2017-07-05
0.4.0

sbellem:


ttmc:
I wonder why there is no analogue of the `__short_version__` seen in bigchaindb/bigchaindb

Issue #315, opened 2017-07-14
Does the Quickstart / Installation Page Work on a new Ubuntu 16.04 VM?


ttmc:
The [current Quickstart / Installation page](https://docs.bigchaindb.com/projects/py-driver/en/v0.4.0/quickstart.html) for the BigchainDB Python Driver only has one OS-level dependency installation:
```text
sudo apt-get update
sudo apt-get install libffi-dev
```
in Step 1. That really doesn't seem like enough, because the `bigchaindb_driver` package depends on the `bigchaindb` package, which needs a lot more OS-level dependencies, i.e. it needs:
```text
sudo apt-get install g++ python3-dev libffi-dev build-essential libssl-dev
```

We should _not_ assume they already have `bigchaindb` (BigchainDB Server) installed. Most people will never run BigchainDB Server.

## To Do

* Test the existing Quickstart / Installation instructions for the Python Driver on a new Ubuntu 16.04 VM.
* Add any needed / missing OS-level dependencies to the Quickstart / Installation page. Re-name that step as "Step 1: OS-level dependencies". It doesn't have to go into the details of what each one is for. You can even cut the existing explanation of the libffi-dev package. It's too much information.

muawiakh:
fixes #316

muawiakh:
No, it does not. Requested a PR

PR #316, opened 2017-07-17
Fix quickstart guide for bigchaindb_driver

muawiakh:
- Fixes quickstart for a fresh Ubuntu/RHEL setup,
and installs all the relevant dependencies.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -5,15 +5,16 @@ Quickstart / Installation
 The BigchainDB Python Driver depends on:
 
 1. ``libffi/ffi.h``
-2. Python 3.5+
-3. A recent Python 3 version of ``pip``
-4. A recent Python 3 version of ``setuptools``
+2. ``libssl-dev``
+3. Python 3.5+
+4. A recent Python 3 version of ``pip``
+5. A recent Python 3 version of ``setuptools``
 
 If you're missing one of those, then see below. Otherwise, you can install the BigchainDB Python Driver (``bigchaindb_driver``) using:
 
 .. code-block:: bash
 
-    pip install bigchaindb_driver
+    $ pip3 install bigchaindb_driver
 
 Next: :doc:`determine the BigchainDB Root URL of the BigchainDB node or cluster you want to connect to <connect>`.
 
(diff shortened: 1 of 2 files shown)


muawiakh:
Fixes #315

docs/quickstart.rst
```diff
@@ -34,20 +35,40 @@ On Ubuntu 14.04 and 16.04, this works:
...
+
+.. code-block:: bash
+
+    $ sudo apt-get install libssl-dev
+
+On Fedora 23 and 24:
+
+.. code-block:: bash
+
+    sudo dnf install openssl-devel
```
ttmc: $

docs/quickstart.rst
```diff
@@ -34,20 +35,40 @@ On Ubuntu 14.04 and 16.04, this works:
...
+
```
ttmc: Below, Python 3.5+ now becomes "Dependency 3". The following ones have to be renumbered as well.

docs/quickstart.rst
```diff
@@ -87,7 +116,12 @@ Once you have a recent Python 3 version of ``pip``, you should be able to upgrad
 
 .. code-block:: bash
 
-    pip install --upgrade setuptools
+    $ pip install --upgrade setuptools
+
+    OR
+
+    $ sudo pip3 install --upgrade setuptools
```
ttmc: Why does this one use sudo but the one above doesn't? Same comment for the next subsection.

ttmc:
Never mind codecov in documentation-only pull requests. Clearly it is wrong.

docs/quickstart.rst
muawiakh: Done

docs/quickstart.rst
muawiakh: Because it didn't work for me without the `sudo` and I was not using the `virtualenv` or `conda`, and running this on base VM. I can change that to `pip3 install --upgrade setuptools` as well to keep it consistent, and if the user encounters a permissions issue they can run it with `sudo` which is pretty straight forward. Same for the next subsection, because with virtualenv, I don't need `sudo` permissions.

muawiakh:
@ttmc because the driver + apt pull the dependencies recursively.

docs/quickstart.rst
ttmc: Okay maybe just remove all the `sudo`

ttmc:
> @ttmc because the driver + apt pull the dependencies recursively.

Okay, so when apt installs `libffi-dev` and `libssl-dev`, it also installs all the other packages. That's nice: only two packages to install.

But why, when we install the `bigchaindb` Python package alone, do we have to install a bunch of other OS-level dependencies? The Python Driver `bigchaindb_driver` package depends on the `bigchaindb` package, so it also depends on all of BigchainDB Server's OS-level dependencies, no?

In short, A depends on B, but A has fewer dependencies than B?

muawiakh:
We only need `libssl-dev` and `libffi-dev` for the server quickstart as well. Nothing more. I will update it.

kremalicious:
So is there a way to have `apt install bigchaindb` automatically install `libssl-dev` and `libffi-dev`? Would be just a bit nicer UX

muawiakh:
@kremalicious No, pip cannot. It can only install python modules as per the `requirements.txt` or `setup.py`

PR #352, opened 2018-01-23
Add BigchainDB server version

vrde:
## Description
I've added the BigchainDB version to the issue template. This can help spotting some incompatibilities fast.


sbellem:
The tests are failing most probably because they run against the bigchaindb master code. Issue https://github.com/bigchaindb/bigchaindb-driver/issues/353 should take care of that.

vrde:
> The tests are failing most probably because they run against the bigchaindb master code. Issue #353 should take care of that.

That's OK. What do you think about the PR?

sbellem:
> That's OK. What do you think about the PR?

It's good.

PR #356, opened 2018-02-27
Problem: 'unique messages for fulfillments' is not included in driver

### LINKED ISSUE #353: Update to latest common code
See https://github.com/bigchaindb/bigchaindb/pull/1937 for background.

codegeschrei:
Solution: update documentation and tests

## Issues This PR Fixes
Fixes #353

## Related PRs
None

## Todos
None

## Deployment Notes
None

## How to QA
None

## Impacted Areas in Application
None

### DIFF
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -5,6 +5,7 @@ import json
 import base58
 from pytest import mark, raises
 from requests.utils import default_headers
+from sha3 import sha3_256
 
 import bigchaindb
 from cryptoconditions import Ed25519Sha256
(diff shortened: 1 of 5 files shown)


tests/test_driver.py
```diff
@@ -98,6 +98,7 @@ def test_prepare(self, driver, alice_pubkey):
         assert not transaction['metadata']
 
     def test_fulfill(self, driver, alice_keypair, unsigned_transaction):
+        from sha3 import sha3_256
```
sbellem: we can import this at the top, as sha3 is an external lib, and is not under test

as an aside sha3 is now in the standard lib for Python 3.6+: https://docs.python.org/3/library/hashlib.html#hash-algorithms

tests/test_offchain.py
```diff
@@ -105,6 +105,7 @@ def test_prepare_transfer_transaction(signed_alice_transaction, recipients):
     ['CT6nWhSyE7dF2znpx3vwXuceSrmeMy9ChBfi9U92HMSP'],
 ))
 def test_fulfill_transaction(alice_transaction, alice_sk):
+    from sha3 import sha3_256
```
sbellem: same as above, we can import this at the top, as sha3 is an external lib, and is not under test

as an aside sha3 is now in the standard lib for Python 3.6+: https://docs.python.org/3/library/hashlib.html#hash-algorithms

PR #369, opened 2018-03-21
Problem: Dev environment is outdated and no integration with Tendermint

### LINKED ISSUE #368: Problem: Dev environment is outdated and no integration with Tendermint
* bigchaindb-driver version: `master`
* bigchaindb **server** version (`bigchaindb -v`): `master/tendermint`
* Python version: `n/a`
* Operating System: `n/a`

### Description
Currently, the development environment of BigchainDB Python driver is outdated and is not integrated with Tendermint i.e. BigchainDB v2.0 is integrated with Tendermint but the driver development enviroment still uses BigchainDB <=1.3.0.

Update the development environment:
- Update `docker-compose.yml`
- Remove `docker-compose.rdb.yml` because RethinkDB is not supported with BigchainDB v2.0
- Update Travis configuration to test with `localmongodb` and BigchainDB v2.0 instead of BigchainDB <=1.3.0 and `rethinkdb`.
- Consolidate all deployment methods in a single `docker-compose` file i.e. `docs.yml` should move into `docker-compose.yml` as well.

muawiakh:
## Description
- Updated `docker-compose.yml`
- Remove `docker-compose.rdb.yml` because RethinkDB is not supported with BigchainDB v2.0
- Update Travis configuration to test with `localmongodb` and BigchainDB v2.0 instead of BigchainDB <=1.3.0 and `rethinkdb`.
- Consolidate all deployment methods in a single `docker-compose` file i.e. `docs.yml` should move into `docker-compose.yml` as well.
- Remove `dev.yml` because it is deprecated(RethinkDB is not supported with 2.0)

## Issues This PR Fixes
Fixes #368 

:fire::fire::fire: Travis will fail with this PR but we will have dependent changes that will take care of those tests @codegeschrei  can confirm :fire::fire::fire:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,29 @@
 language: python
 python:
+  - 3.5
   - 3.6
 
 env:
   global:
     - DOCKER_COMPOSE_VERSION=1.19.0
   matrix:
+    - TOXENV=py35
     - TOXENV=py36
     - TOXENV=flake8
     - TOXENV=docs
-  
+
+matrix:
+  fast_finish: true
+  exclude:
+    - python: 3.5
+      env: TOXENV=flake8
+    - python: 3.5
+      env: TOXENV=docs
+    - python: 3.5
+      env: TOXENV=py36
+    - python: 3.6
+      env: TOXENV=py35
+
 before_install:
   - sudo .ci/travis-before-install.sh
 
--- a/.ci/travis-after-success.sh
+++ b/.ci/travis-after-success.sh
@@ -2,6 +2,6 @@
 
 set -e -x
 
-if [ "${TOXENV}" == "py36" ]; then
+if [ "${TOXENV}" == "py35" || "${TOXENV}" == "py36" ]; then
     codecov -v
 fi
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -1,26 +1,15 @@
...
-  - TOXENV=flake8
-  - TOXENV=docs
+  global:
+    - DOCKER_COMPOSE_VERSION=1.19.0
+  matrix:
+    - TOXENV=py36
+    - TOXENV=flake8
+    - TOXENV=docs
   
-matrix:
```
shahbazn: un-related to this PR by why we dont support python 3.5 now? because thats still the default for ubuntu-xenial and other latest OSes I assume? cc @vrde

vrde: We support only Python 3.6. I quickly talked with @muawiakh about that. We can drop `tox` for now IMO.

muawiakh: Yes, I will remove it from this PR. If we see a problem with the driver while testing with `3.5` we can deprecate the support. I will revert this change for now.

codegeschrei:
I think the basics are working for now. We might encounter more problems later but that's why I also created a branch to collect all changes for bdb-2.0, so we can fix everything step by step before it goes into master.

PR #366, opened 2018-03-13
Problem: block_id parameter for block query is now block_height

codegeschrei:
Solution: change parameter to block_height and update documentation

## Issues This PR Fixes
Fixes #362 

## Related PRs
bigchaindb/bigchaindb#1970

## Todo

- [ ] fixing the tests, as soon as we have a new dev environment

PR #379, opened 2018-03-27
Problem: Transaction cannot be present in multiple blocks

kansi:
Solution: Migtrate blocks api to return one block id or None

Refer  #366

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -124,15 +124,14 @@ block.
 .. code-block:: python
 
     # Retrieve a block height
-    >>> block_height = bdb.transactions.get(txid=signed_tx['id'])
+    >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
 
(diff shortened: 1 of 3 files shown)


docs/advanced-usage.rst
```diff
@@ -124,15 +124,14 @@ block.
...
 
-This will return a list with block heights. If the transaction is not in any block an empty list will be returned.
-If we want some more data we can use the block height to retrieve the block itself.
+This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is returned. If we want some more data we can use the block height to retrieve the block itself.
 
 .. code-block:: python
 
     # Retrieve a block
-    >>> block = bdb.transactions.retrieve(str(block_height[0]))
+    >>> block = bdb.blocks.retrieve(block_height[0])
```
codegeschrei: `block_height[0]` this should be just `block_height` now, right?

PR #382, opened 2018-03-29
bigchaindb-driver requires python 3

NickFitton:
## Description
noticed the readme reports to use `pip`, `bigchaindb_driver` requires python 3 to install so it using `pip3` works.

## Issues This PR Fixes
Fixes misleading documentation (No `#` issue).


codegeschrei:
Hey! Thanks for the PR and your input.
In general we recommend people to use a virtualenv and within you can just use `pip` which would be misleading in this context.
Maybe we can change it either providing both versions and hint to a virtualenv for just `pip` or just add a hint matching the one version we provide, what do you think?

NickFitton:
Sounds like an idea, I'll make the changes.

codegeschrei:
Hey! Can you merge master. The tests should be passing afterwards and I can merge it :)

NickFitton:
:ok_hand:

PR #393, opened 2018-04-06
Problem: docs are not building with python 3.5

codegeschrei:
Solution: change to python 3.6

To change the version from 3.5 to 3.6 a config file for readthedocs needs to be added that specifies the wanted version

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+build:
+    image: latest
+
+python:
+    version: 3.6
+    pip_install: true


codegeschrei:
I like 2 but this needs some more time, so maybe we can go for 1?
We can still, in the future, try 2.

ttmc:
I would be okay with solution 1. if the use of "f strings" is the _only_ reason we need Python 3.6+.

vrde:
I'll close this PR since the problem has been temporary fixed in the Server code.

PR #405, opened 2018-04-19
Problem: make start should only start server containers

codegeschrei:
Solution: remove container to run tests

We already have a command to run tests, so they should not be run with `make start`


shahbazn:
Changes make sense, Can you please look into why the tests are failing

codegeschrei:
@shahbazn I restarted it. This seems to be a randomly failing test, it happened before too. I need to investigate it later.

PR #410, opened 2018-04-23
Problem: no specific python version for pip

### LINKED ISSUE #350: Set python_requires >=3.5 in setup.py
https://packaging.python.org/tutorials/distributing-packages/#python-requires

codegeschrei:
Solution: set python version for pypi

## Issues This PR Fixes
Fixes #350

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
                  'bigchaindb_driver'},
     include_package_data=True,
     install_requires=install_requires,
+    python_requires='>=3.5',
     license="Apache Software License 2.0",
     zip_safe=False,
     keywords='bigchaindb_driver',


ttmc:
For reference, here's the documentation about `python_requires`: https://packaging.python.org/tutorials/distributing-packages/#python-requires

> Note: Support for this feature is relatively recent. Your project’s source distributions and wheels (see Packaging your project) must be built using at least version 24.2.0 of setuptools in order for the python_requires argument to be recognized and the appropriate metadata generated.
In addition, only versions 9.0.0 and higher of pip recognize the python_requires metadata. Users with earlier versions of pip will be able to download & install projects on any Python version regardless of the projects’ python_requires values.

PR #411, opened 2018-04-25
Problem: the driver depends on BigchainDB package

vrde:
Solution: copy the module `common` in the source of the driver, and update dependencies.

Copy code is not the ideal solution, but we need to unlock https://github.com/bigchaindb/bigchaindb/pull/2235.

Moreover, we have a [path forward](https://github.com/bigchaindb/BEPs/pull/33).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,10 @@ with open('CHANGELOG.rst') as changelog_file:
 
 install_requires = [
     'requests>=2.11.0',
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -11,7 +11,10 @@
 
 install_requires = [
     'requests>=2.11.0',
-    'bigchaindb~=2.0.0a2',
```
kansi: :smiley:

PR #412, opened 2018-05-07
Problem: the driver is using an old tendermint version

codegeschrei:
Solution: upgrade tendermint to match the version the server uses

The server recently upgraded to a newer tendermint version and these changes were not yet reflected in the driver. They also led to problems when using docker.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,14 +51,14 @@ services:
       retries: 3
     command: bigchaindb -l DEBUG start
   tendermint:
-    image: tendermint/tendermint:0.12
+    image: tendermint/tendermint:0.19.2
     volumes:
       - ./compose/tendermint/tmdata/config.toml:/tendermint/config.toml
     entrypoint: ''
     ports:
       - "46656"
       - "46657"
-    command: bash -c "tendermint init && tendermint node"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -51,14 +51,14 @@ services:
...
-    image: tendermint/tendermint:0.12
+    image: tendermint/tendermint:0.19.2
     volumes:
       - ./compose/tendermint/tmdata/config.toml:/tendermint/config.toml
     entrypoint: ''
     ports:
       - "46656"
       - "46657"
-    command: bash -c "tendermint init && tendermint node"
+    command: sh -c "tendermint init && tendermint node --proxy_app=tcp://bigchaindb:46658"
```
kansi: @codegeschrei You can also [disable empty blocks](https://github.com/bigchaindb/bigchaindb/blob/master/docker-compose.yml#L56)

codegeschrei: oh right! good idea! and thanks for the link

vrde: We recently added `--consensus.create_empty_blocks=false` to Tendermint: https://github.com/bigchaindb/bigchaindb/blob/master/docker-compose.yml#L56

PR #414, opened 2018-05-07
Release for 0.5 Alpha 3

codegeschrei:


vrde:
Should we mention that BigchainDB Driver doesn't require BigchainDB anymore?

Issue #419, opened 2018-05-11
transaction status call missing from the latest driver


agwego:
* bigchaindb-driver version: 0.5.0a4
* bigchaindb **server** version (`bigchaindb -v`): branch alpha-3
* Python version: 3.5.2
* Operating System: Ubuntu 16.04.3 LTS

The trnasaction.status call is missing from the driver.

vrde:
Hi @agwego, when you send a transaction to the network, the driver will wait for that transaction to be included in a block (this is true since version `0.5.0a2`). If there is a problem, than the call will raise an exception.

I'll keep this issue open so we remember to be more explicit in the documentation and examples.

agwego:
You're missing the point.

I have code in production against BigchainDB 1.3, I'm in the process of migrating that code to 2.X, I now have to make the decision do I branch my code which is annoying, or work around the issue, because you changed the API and now my code doesn't work with 2.X. This includes Python, Java, Node.js code bases. I've tweaked your codebase quite a bit, because your logging leaves a lot to be desired.

codegeschrei:
Hey @agwego I've opened a PR to show how to check for the status of a transaction. If you're interested you can give some feedback.

PR #421, opened 2018-05-16
Problem: It is not known how to check the status of a transaction

codegeschrei:
Solution: explain how to check for the status of a tx in the docs

related to #419

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -127,8 +127,8 @@ block.
     >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
 
 This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is
-returned. If it is ``None`` it can have different reasons for example the transaction was not valid or is
-still in the queue and you can try again later. If the transaction was invalid or could not be sent an exception is raised.
+returned. If it is ``None`` it can have different reasons for example the transaction was invalid or is still in the
+queue and you can try again later. If the transaction was invalid or could not be sent an exception is raised.
 
 If we want to see the whole block we can use the block height to retrieve the block itself.
 
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -115,23 +115,27 @@ Recap: Asset Creation
     sent_tx = bdb.transactions.send(signed_tx)
     sent_tx == signed_tx
 
-Read the Creation Transaction from the DB
------------------------------------------
+Check if the Transaction was sent successfully
+-----------------------------------------------
 
-After a couple of seconds, we can check if the transaction was validated in a
+After a couple of seconds, we can check if the transaction was included in a
 block.
 
 .. code-block:: python
 
-    # Retrieve a block height
+    # Retrieve the block height
     >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
 
-This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is returned. If we want some more data we can use the block height to retrieve the block itself.
+This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is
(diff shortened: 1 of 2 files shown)


docs/usage.rst
```diff
@@ -140,6 +140,28 @@ Notice the transaction ``id``:
 
     In [0]: txid
 
+Check if the Transaction was sent successfully
+-----------------------------------------------
+
+After a couple of seconds, we can check if the transaction was validated in a
```
ttmc: I would write "included" rather than "validated" here. I realize you were just copying text from before, but we may as well fix it now.

docs/usage.rst
```diff
@@ -140,6 +140,28 @@ Notice the transaction ``id``:
...
+block.
+
+.. code-block:: python
+
+    # Retrieve the block height
+    >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
+
+This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is
+returned. If it is ``None`` it can have different reasons for example the transaction was not valid or is
+still in the queue and you can try again later. If the transaction was invalid or could not be send an error message
```
ttmc: or could not be _sent_

docs/usage.rst
codegeschrei: yes, that makes more sense!

docs/advanced-usage.rst
```diff
@@ -115,23 +115,27 @@ Recap: Asset Creation
...
 
 .. code-block:: python
 
-    # Retrieve a block height
+    # Retrieve the block height
     >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
 
-This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is returned. If we want some more data we can use the block height to retrieve the block itself.
+This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is
+returned. If it is ``None`` it can have different reasons for example the transaction was not valid or is
```
z-bowen: In line 130 you write 'the transaction was not valid' and in line 131 you say 'was invalid'. Unless these have some semantic difference, I'd suggest using one consistent word choice everywhere, i.e. 'invalid' instead of 'not valid'

docs/advanced-usage.rst
```diff
@@ -115,23 +115,27 @@ Recap: Asset Creation
...
+still in the queue and you can try again later. If the transaction was invalid or could not be sent an exception is raised.
+
+If we want to see the whole block we can use the block height to retrieve the block itself.
 
 .. code-block:: python
 
-    # Retrieve a block
-    >>> block = bdb.blocks.retrieve(block_height)
+    # Retrieve the block that contains the transaction
+    >>> block = bdb.blocks.retrieve(str(block_height))
```
z-bowen: Why do we need to type-cast `block_height` as a string? Shouldn't this be an `int`, and shouldn't `.retrieve` be smart enough to duck-type it?

docs/usage.rst
```diff
@@ -140,6 +140,27 @@ Notice the transaction ``id``:
...
+block.
+
+.. code-block:: python
+
+    # Retrieve the block height
+    >>> block_height = bdb.blocks.get(txid=signed_tx['id'])
+
+This will return the block height containing the transaction. If the transaction is not in any block then ``None`` is
+returned. If it is ``None`` it can have different reasons for example the transaction was not valid or is
+still in the queue and you can try again later. If the transaction was invalid or could not be sent an exception is raised.
```
z-bowen: This line is breaking the 119 character limit.

z-bowen:
Made some small suggestions, but nothing that requires a second review

docs/advanced-usage.rst
codegeschrei: You can open an issue for it if you want.

PR #423, opened 2018-05-16
Problem: The error message for Double Spends changed

codegeschrei:
Solution: change the Double Spends error message in the docs


docs/advanced-usage.rst
```diff
@@ -347,7 +347,8 @@ Fulfill the transaction
        ...:     private_keys=alice.private_key,
        ...: )
 
-Send the transaction over to the node
+Send the transaction over to the node and an error will occur, informing about the double spend and
```
z-bowen: Writing 'informing the user of the double spend' is better English.

ttmc:
@z-bowen I think @ldmberman is away until next Tuesday, so if you're okay with this PR, you can merge it.

PR #428, opened 2018-05-18
Problem: the updated Makefile is not mentioned in the docs

### LINKED ISSUE #425: Problem: Docs about getting Root URL of local Docker-based node are out-of-date
In the Python Driver docs page "[Determine the BigchainDB Root URL](https://docs.bigchaindb.com/projects/py-driver/en/latest/connect.html)", the last section is about how to determine the Root URL when running a local Docker-based node.

That section is out-of-date, referring to some old things, including another section in the docs:

https://docs.bigchaindb.com/projects/py-driver/en/latest/contributing.html#devenv-docker

The modern way to develop using Docker is to use the `make` commands in the BigchainDB Makefile, so the above-mentioned two sections in the docs should be updated to explain how to use that instead.

codegeschrei:
Solution: add useful Makefile commands to the docs

Fixes #425

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ release: dist ## package and upload a release
 	twine upload dist/*
 
 root-url:
-	@docker-compose port bigchaindb 9984
+	@$(DC) port bigchaindb 9984
 
 ###############
 # Sub targets #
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -84,6 +84,9 @@ clean: clean-build clean-pyc clean-test ## Remove all build, test, coverage and
 release: dist ## package and upload a release
 	twine upload dist/*
 
+root-url:
+	@docker-compose port bigchaindb 9984
```
ttmc: I wonder why you used `@docker-compose` here rather than `@$(DC)`

codegeschrei: oh! I forgot to change it back after testing something. Will fix it

ttmc:
Travis is failing when it tries to build the docs.

codegeschrei:
It should actually be bash instead... guess I need another coffee D:

codegeschrei:
I think I tried `none` once before and it's not working bc it's a required argument
I'll fix it later and change it to bash

PR #430, opened 2018-05-28
BEP 14: Roundrobin connection strategy

tucanae47:
# Description
Added support for [BEP/14](https://github.com/bigchaindb/BEPs/tree/master/14) to improve Drivers Reliability 
Added new test set for mutiple nodes 

The approach implemented is the roundrobin connection strategy previously discussed on  [bep14 issue](https://github.com/bigchaindb/BEPs/issues/40)

## How to QA
For running tests with docker i used suggested approach on contributing guideline command with an addition to test multiple nodes:
```
$ docker-compose up -d bigchaindb
#add line to run multiple bigchain nodes on the same docker network
$ docker-compose up -d --scale bigchaindb=5
$ docker-compose run --rm bigchaindb-driver pytest -v tests/test_driver.py
```


bigchaindb_driver/pool.py
```diff
@@ -46,9 +52,16 @@ def pick(self, connections):
                 :class:`~bigchaindb_driver.connection.Connection` instances.
 
         """
-        self.picked += 1
-        self.picked = self.picked % len(connections)
-        return connections[self.picked]
+
+        self.current_time_ms = datetime.now()
```
vrde: `current_time_ms` can be a local variable, I guess it's safe to remove `self` (and remove it form the `__init__`).

bigchaindb_driver/utils.py
```diff
@@ -60,20 +59,44 @@ def _get_default_port(scheme):
...
-
+        url = _normalize_url(node)
+        norm_nodes.append(url)
     return tuple(norm_nodes,)
+
+
+def _normalize_nodes(*nodes):
+    if not nodes:
+        return (DEFAULT_NODE,)
+    if isinstance(nodes[0], dict):
```
vrde: The best practice is to use `issubclass`, so in case someone is passing a custom object that inherits from `dict`, this check will still work. (I doubt this will ever create problems, but _safe is better than sorry_.)

bigchaindb_driver/utils.py
```diff
@@ -60,20 +59,44 @@ def _get_default_port(scheme):
...
+        return _normalize_dict(*nodes)
```
vrde: I see that both functions `_normalize_dict` and `_normalize_array` accept a `*nodes`, and when you are calling them you are unpacking `nodes` as well with the star operator. Would it make sense to just pass the object, without unpacking, and adjust the function signature as well? I feel unpacking is a bit redundant in this case.

tests/test_driver.py
```diff
@@ -6,8 +6,8 @@
 from pytest import mark, raises
 from requests.utils import default_headers
 from sha3 import sha3_256
-
 from cryptoconditions import Ed25519Sha256
+import asyncio
```
vrde: I see the only function used from the `asyncio` module is `asyncio.sleep`. Why not using `time.sleep`? It would simplify the tests.

codegeschrei:
hey, one general thing I noticed: some doc strings have changed. Can you check your changes and update the doc strings accordingly?

tests/test_driver.py
```diff
@@ -223,6 +291,47 @@ def test_retrieve(self, driver, block_with_alice_transaction):
         assert block
 
 
+class TestBlocksEndppointMultiple:
```
codegeschrei: there is a typo here `Endpoint` with two p

tests/test_driver.py
```diff
@@ -223,6 +291,47 @@ def test_retrieve(self, driver, block_with_alice_transaction):
...
+
+    @mark.asyncio
+    @mark.parametrize("timeout", [0.5])
+    async def test_get_timeout(self, timeout, event_loop,
```
codegeschrei: the name of this (and the tests below) are a bit confusing because you don't get/receive a timeout/loop

tests/test_pool.py
```diff
@@ -1,9 +1,13 @@
 def test_get_connection():
+    from datetime import datetime
     from bigchaindb_driver.pool import Pool
-    pool = Pool((0, 1))
+    connections = [{"node":0, "time":datetime.now()}, {"node": 1, "time":datetime.now()}]
```
codegeschrei: there are some whitespaces missing after `:`

tests/conftest.py
```diff
@@ -11,12 +11,14 @@
 from pytest import fixture
 from sha3 import sha3_256
 
-from bigchaindb_driver.common.transaction import Transaction, _fulfillment_to_details
-
+from bigchaindb_driver.common.transaction import _fulfillment_to_details
+from bigchaindb_driver.common.transaction import Transaction
```
codegeschrei: Why did you split the imports here?

tests/conftest.py
```diff
@@ -176,6 +178,48 @@ def bdb_node(bdb_host, bdb_port):
...
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_4:9984',
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_5:9984',
+             'headers': {'app_id': 'id'}},
+            ]
+
+
+@fixture
+def driver_multiple(bdb_nodes):
```
codegeschrei: where are we using this fixture?

tests/test_driver.py
```diff
@@ -19,6 +19,15 @@ class TestBigchainDB:
...
                                                   'http://node-2:9984')),
+        (({'endpoint': 'node-1',
+           'headers': {'app_id': 'id'}},
+          {'endpoint': 'node-2',
+           'headers': {'app_id': 'id'}}),
+            None,
+            ({'endpoint': 'http://node-1:9984',
+              'headers': {'app_id': 'id'}},
+             {'endpoint': 'http://node-2:9984',
+              'headers': {'app_id': 'id'}})),
```
codegeschrei: would it make sense to just remove line 18/19 because it is more or less the same set of nodes?

tests/conftest.py
```diff
@@ -176,6 +178,48 @@ def bdb_node(bdb_host, bdb_port):
...
+            {'endpoint': 'bigchaindb-driver_bigchaindb_1:9984',
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_2:9984',
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_3:9984',
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_4:9984',
+             'headers': {'app_id': 'id'}},
+            {'endpoint': 'bigchaindb-driver_bigchaindb_5:9984',
+             'headers': {'app_id': 'id'}},
```
codegeschrei: would it make sense to have a fixture or var for the `header` since it doesn't change?

tests/test_driver.py
tucanae47: the main reason why i included ```asyncio```, is to make use of its features on future multiple nodes tests and the following ```FIXME``` comment in conftest.py, line 16: 
```(python)
# FIXME The sleep, or some other approach is required to wait for the
# transaction to be available as some processing is being done by the
# server.
```
let me know if it makes sense, otherwise i'll remove it.

tucanae47:
Thank you guys for the review, I'm pushing a new commit with the suggested changes.I agree with mostly everything, so i updated the implementation accordingly.

tests/conftest.py
tucanae47: it raised a flake8 warning, so i updated it with a parenthesis approach for multiple imports

codegeschrei:
Hey, did you install `pre-commit`? in `transport.py` some checks failed for me.

tucanae47:
no i didn't sorry, just installed and fixed accordingly.

bigchaindb_driver/pool.py
```diff
@@ -73,7 +104,9 @@ def get_connection(self):
             A :class:`~bigchaindb_driver.connection.Connection` instance.
 
         """
-        if len(self.connections) > 1:
-            return self.picker.pick(self.connections)
 
-        return self.connections[0]
+        if self.tries >= self.max_tries:
```
vrde: `tries` is incremented every time there is a failure and when it reaches the limit, but never decremented, so if someone is using the driver for enough time to accumulate enough errors, `get_connection` will always return `None`.

In general, I think it's difficult to use a counter to decide if the network is unresponsive or not. Let's say you have 4 nodes, and one of them is always failing (but the other 3 are fine). After some requests, `tries` will reach `max_tries` and `get_connection` will return you `None`, even if the network is responsive.

An idea would be to count how many nodes have been `DELAY`ed, and if all nodes were unavailable, then raise an exception and reset the nodes `DELAY`s. What do you think?

bigchaindb_driver/transport.py
```diff
@@ -50,11 +74,28 @@ def forward_request(self, method, path=None,
...
+                self.pool.fail_node()
+                return self.forward_request(
+                    method=method,
+                    path=path,
+                    params=params,
+                    json=json,
+                    headers=headers,
+                    error=err)
+        elif error is not None:
+            raise error
```
vrde: What do you think is better? Raising `err` (that is _the last error_) or `error` (that is _the first error_)?

bigchaindb_driver/pool.py
tucanae47: I Agree. so i updated the implementation, using:

1. A new dict ```retries_left``` per node, to keep track of retries left per node. 
2. Updating ```retries_left``` based not only when node is failing but also when node is successful. 
3. and Return ```None # Service Unavailable``` when  ```retries_left per node == 0``` for all nodes.

bigchaindb_driver/transport.py
```diff
@@ -50,11 +74,28 @@ def forward_request(self, method, path=None,
...
+                response = connection.request(
+                    method=method,
+                    path=path,
+                    params=params,
+                    json=json,
+                    headers=headers,
+                )
+                self.pool.success_node()
+                return response.data
+            except BaseException as err:
```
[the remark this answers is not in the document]
tucanae47: I Agree, but i used ```BaseException``` as in ```Exception``` cases which are not ```except:```ed in ```TransportError```  (Timeout, MaxRetries) . Will raise for example:
```
requests.exceptions.ConnectionError: 
HTTPConnectionPool(host='bigchaindb-driver_bigchaindb_2', port=9984): 
Max retries exceeded with url: /api/v1/blocks/69 
(Caused by NewConnectionError('<urllib3.connection.HTTPConnection 
objectat 0x7f2be1d815c0>:
Failed to establish a new connection: [Errno -2] Name or service not known',))
```
which well, im not sure yet how to handle those better. What do you think?

bigchaindb_driver/transport.py
tucanae47: We could not raise```err``` as at that point is not available locally and it will throw
```
UnboundLocalError: local variable 'err' referenced before assignment
```

bigchaindb_driver/pool.py
tucanae47: Couple of things to clarify item 2 and 3.
> 2. Updating retries_left based not only when node is failing but also when node is successful.

```retries_left``` per node is based on max and min constraints 
```
min_retries = 0
max_retries = len(nodes)*4. #could be another max limit
```

> 3. and Return None # Service Unavailable when retries_left per node == 0 for all nodes.

I think there is a problem particularly if the list of nodes is very big. Lets consider a worst case scenario:
50 nodes: 1 is ok, 49 failing. So the round_robin will try the 49 nodes after reaching the correct one again. That suggests a limit to the number of nodes, what do you think?

File: bigchaindb_driver/pool.py [hunk 15]
```diff
@@ -46,9 +51,16 @@ def pick(self, connections):
...
+
+        current_time_ms = datetime.now()
+        picked_time = connections[self.picked]["time"]
+        if current_time_ms > picked_time:
+            node = connections[self.picked]["node"]
+            return node
+        else:
+            self.next_node(connections)
+            node = connections[self.picked]["node"]
+            return node
```
kansi: One way to simplify this conditional would be as follows,

```python
       current_time_ms = datetime.now()
       picked_time = connections[self.picked]["time"]
       if current_time_ms < picked_time:
            self.next_node(connections)

       node = connections[self.picked]["node"]
       return node
```

kansi:
This PR looks good but I have one concern though,

`self.get_connection()` decides if nodes are available only if there are retires left for that node i.e.

```python
        for node in self.retries:
            if self.retries[node] > 0:
                return True
        return False
```

and the next node is picked if the existing node has been put into delay i.e. 

```python
        if current_time_ms > picked_time:
            node = connections[self.picked]["node"]
            return node
        else:
            self.next_node(connections)
            node = connections[self.picked]["node"]
            return node
```

So, for a network of size, say 10 nodes, the `max_tries` for each node would be [`40`](https://github.com/bigchaindb/bigchaindb-driver/pull/430/files#diff-b077e6085190e2851885296fb94d95e7R79). Since the available nodes are only decided based on their available retries a net total of `400` retries will be attempted for a given failing request before the driver gives up and raises an exception.

Two questions:
- Is my understanding right?
- Can this be an issue?

/cc @vrde

bigchaindb_driver/pool.py
```diff
@@ -46,9 +51,16 @@ def pick(self, connections):
                 :class:`~bigchaindb_driver.connection.Connection` instances.
 
         """
-        self.picked += 1
-        self.picked = self.picked % len(connections)
-        return connections[self.picked]
+
```
ldmberman: In `pick` you only check for the node delay once - if the first connection is not available you return the second one without checking if it is available as well. This way at some point the driver will make requests to the previously failed nodes without respecting the delays.

Instead, `pick` should return the first available connection, somthing like this:

```
cursor = self.picked
while cursor < self.picked + len(connections):
    if current_time_ms > connections[cursor]['time]:
        self.picked = cursor // len(connections)
        return cursor
    
    cursor += 1
```

bigchaindb_driver/pool.py
```diff
@@ -63,7 +75,45 @@ def __init__(self, connections, picker_class=RoundRobinPicker):
 
         """
         self.connections = connections
+        self.node_count = len(self.connections)
+        self.max_retries = self.node_count * 4
```
ldmberman: Fix me if I am wrong but it looks like `get_connection` will forever return `None` after the retry maximum is reached..

You can remove the cap and instead add [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) - in the simplest form it would look like:

```
delay = initial_delay * 2 ** retries
```
This will allow the driver users to wait for a successful connection for as long as they wish while optimizing the load put onto the network.

Also, makes sense to start with a small delay, some fraction of a second. The client should not wait for a minute if it's an intermittent network failure.

bigchaindb_driver/pool.py
```diff
@@ -63,7 +75,45 @@ def __init__(self, connections, picker_class=RoundRobinPicker):
...
+                return True
+        return False
+
+    def success_node(self):
+        """Send a message to the pool indicating the connection to the current
+        node is succesful
+        """
+        success_node = self.picker.picked
+        self.retries_left(success_node, success=True)
+        self.picker.next_node(self.connections)
```
ldmberman: I think, we should stay to exploit the node if the request is successful. If the node is capable of serving so many requests the client request times will be improved significantly. If the node can't handle it, we back off. Also note that each driver user assembles the list of the nodes by himself so the initial distribution should be close to uniform.

bigchaindb_driver/pool.py
```diff
@@ -73,7 +123,9 @@ def get_connection(self):
...
         """
-        if len(self.connections) > 1:
-            return self.picker.pick(self.connections)
 
-        return self.connections[0]
+        if not self.nodes_available():
+            return None
+        elif len(self.connections) > 1:
+            return self.picker.pick(self.connections)
+        return self.connections[0]["node"]
```
ldmberman: `pick` should deal with the special case of a single connection in the list. It makes sense to retry requests even when there is a single node configured.

bigchaindb_driver/transport.py
```diff
@@ -21,8 +23,30 @@ def __init__(self, *nodes, headers=None):
 
     def init_pool(self, nodes, headers=None):
         """Initializes the pool of connections."""
+        if isinstance(nodes[0], dict):
+            self.init_nodes_dict(nodes)
```
ldmberman: If `nodes` are passed as a dictionary, the `headers` argument is ignored, what might be confusing..

It might be hard to maintain multiple ways of describing nodes so I tried to come up with an alternative that uses a single input format, yet preserving backwards-compatibility. What do you think about it? 

```
def init_pool(self, nodes, headers=None, node_specific_headers=None):
    """Initializes the pool of connections."""
    
    self.pool = Pool([Connection(node_url=node, headers=self._get_node_headers(node))
                      for node in nodes])

def _get_node_headers(self, node, common_headers=None, node_specific_headers=None):
    if common_headers is None:
        common_headers = {}
    if node_specific_headers is None:
        node_specific_headers = {}
    return {**common_headers, **node_specific_headers.get(node, {})}
```

bigchaindb_driver/pool.py
kansi: How should this be reconciled with the [exponential back-off](https://github.com/bigchaindb/bigchaindb-driver/pull/430#discussion_r194366351) i.e. In case of a single node should that back-off strategy be followed? 

- If no, then what should be the alternate approach?
- If yes, then following are the retry intervals (with `initial_delay = 1`): `2`, `4`, `8`, `16`. In this case should the client wait for all the attempts i.e. `16` seconds?

bigchaindb_driver/transport.py
```diff
@@ -50,11 +74,28 @@ def forward_request(self, method, path=None,
 
         """
         connection = self.get_connection()
-        response = connection.request(
```
ldmberman: We need to accept a timeout here - driver users should be able to specify for how long they are able to wait for the response. Consequently, a timeout has to be specified for each HTTP request.

Here are some code snippets which I hope are helpful:
```
time_left = datetime.timedelta(timeout)
start = time.time()
while True:
    connection = self.get_connection()

    if connection is not None:
        response = self._request(..timeout=time_left.seconds)
        if response is not None:
            return response
    
    end = time.time()
    time_left -= datetime.timedelta(seconds=int(end-start))
    if time_left.total_seconds() <= 0:
        raise ConnectionTimeoutError  # or smth similar
```
`def _request`:

```
try:
    response = connection.request(..timeout=timeout)
except Exception as e:
    log.warning(f'An exception occurred while connecting to the node {path}: {e}')
    self.pool.fail_node()
    return
else:
    self.pool.success_node()
    return response.data
```
Note the logging - you can instantiate a logger in the global scope via `log = logging.getLogger(__name__)`.

Also, I simplified error handing in these snippets - please, tell me if I missed anything or if there are any other concerns to take into account. For instance, what is the purpose of passing `error` to `forward_request`?

tests/test_driver.py
ldmberman: You do not need to sleep at all. Transactions are now posted [in the commit mode](https://github.com/bigchaindb/bigchaindb-driver/blob/master/tests/conftest.py#L193) so when you get the response the data is stored and available by definition.

The `await_transaction` fixture (you referred to its FIXME) is a confusing piece of legacy code, that is going to be removed soon.

bigchaindb_driver/pool.py
ldmberman: Yes, it should follow the same strategy when there is only a single node.

The client should wait for as long as it wants, what can be expressed via an optional timeout parameter. Some more details [here](https://github.com/bigchaindb/bigchaindb-driver/pull/430/files#r194404305).

bigchaindb_driver/offchain.py
```diff
@@ -179,7 +179,7 @@ def prepare_create_transaction(*,
     if not recipients:
         recipients = [(signers, 1)]
     elif not isinstance(recipients, (list, tuple)):
-            recipients = [([recipients], 1)]
+        recipients = [([recipients], 1)]
```
muawiakh: :thumbsup:

bigchaindb_driver/pool.py
```diff
@@ -63,7 +75,45 @@ def __init__(self, connections, picker_class=RoundRobinPicker):
 
         """
         self.connections = connections
+        self.node_count = len(self.connections)
+        self.max_retries = self.node_count * 4
+        self.retries = {
+            key: self.max_retries for key in range(self.node_count)
+        }
         self.picker = picker_class()
+        self.DELAY = 60
```
muawiakh: 60 seconds for a default delay? Any particular reason? I like what Lev proposed by adding exponential back off and starting off with a small delay.

bigchaindb_driver/pool.py
tucanae47: i agree, so i updated accordingly with a minor change: 
```
   while current_time_ms <= connections[self.picked]['time']:
        self.next_node(connections)
        current_time_ms = datetime.utcnow()
   return connections[self.picked]
```
```next_node``` keeps iterating over complete set of nodes, until time during backoff is exhausted and we find an available node. please check last push.

File: bigchaindb_driver/pool.py (hunk 17 above)
Comment (tucanae47) [reply bigchaindb_driver/pool.py]: i agree and you were not wrong, please check next push.

File: bigchaindb_driver/transport.py (hunk 20 above)
Comment (tucanae47) [reply bigchaindb_driver/transport.py]: to be honest i don't understand what you propose as improvement, and well the signature kind of disagrees with the [BEP14](https://github.com/bigchaindb/BEPs/tree/master/14) signature proposal, because of the new parameter your introduced for adding specific node headers.

on the other hand, i disagree about the current approach to be confusing. Any ways, care to elaborate more?

File: bigchaindb_driver/transport.py (hunk 21 above)
Comment (tucanae47) [reply bigchaindb_driver/transport.py]: thanks for the snippets, implemented on the next push with a equivalent mindset, and updated from recursive to iterative approach. 
also about:
> Also, I simplified error handing in these snippets - please, tell me if I missed anything or if there are any other concerns to take into account. For instance, what is the purpose of passing error to forward_request?

is the last error we obtained from the set of requests executed during the loop, some tests will not pass if we raise timeout exception for all posible exceptions raised, so the last exception raised will have more priority over the timeout, and the goal is to inform (give a hint) the user what is wrong with the last node used.

File: bigchaindb_driver/pool.py (hunk 23 above)
Comment (tucanae47) [reply bigchaindb_driver/pool.py]: Is the default timeout for [tcp_fin_timeout](http://man7.org/linux/man-pages/man7/tcp.7.html)  for FIN packets on linux tcp sockets. which i considered a good constant for this case. any way i updated accordingly to @ldmberman. next push.

**Comment by tucanae47** [comment]
Test are failing as stated by @ttmc on this [comment](https://github.com/bigchaindb/bigchaindb-driver/pull/443#issuecomment-397007356)

**Comment by codegeschrei** [comment]
@tucanae47 I fixed the tests. You can just merge master and it should work :)

File: bigchaindb_driver/transport.py (hunk 20 above)
Comment (ldmberman) [reply bigchaindb_driver/transport.py]: I find it harder to figure out how the function works when 2 different input formats are involved. For example, by looking at the function signature, a user may infer that the function can accept some nodes as plain strings and others as dictionaries. May be he has 10 nodes and wants to specify a custom header for only one of them so he may try to do `init_pool(nodeurl1, nodeurl2, ... nodeurl9, {'node': nodeurl10, 'headers': ..})` what will fail in some weird way w.r.t current implementation. Another point here is if I only want custom headers for a small subset of nodes it would be easier for me to only alter the way I pass this node as a parameter, not all nodes.

What do you think about the following alternative, still using 2 formats, but allowing to only specify a dict when it's really needed?
```
def init_pool(nodes, headers=None):
    connections = []
    for node in nodes:
        if isinstance(node, str):
           connections.append({'node': Connection(node_url=node, headers=headers), 'time': time})
        else:
           node_headers = self._merge_headers(node['headers'], headers)
           connections.append({'node': Connection(node_url=node['endpoint'], headers=node_headers), 'time': time})
```

File: bigchaindb_driver/pool.py (hunk 23 above)
Comment (ldmberman) [reply bigchaindb_driver/pool.py]: @tucanae47 `tcp_fin_timeout` is about how much to wait before shutting the connection down - more like the timeout the user will pass to `forward_request` in this implementation, apparently it's different from the backoff delay.

**Comment by ldmberman** [comment]
> so the last exception raised will have more priority over the timeout, and the goal is to inform (give a hint) the user what is wrong with the last node used

@tucanae47 that's a good point! What do you think about collecting all the errors that occurred? It should be even more helpful.

Also, when the global timeout expires, I would append a custom timeout error to the list instead of mimicking an actual HTTP error cause in this case it's not an HTTP error obviously.

File: bigchaindb_driver/pool.py [hunk 24]
```diff
@@ -46,8 +51,13 @@ def pick(self, connections):
                 :class:`~bigchaindb_driver.connection.Connection` instances.
 
         """
-        self.picked += 1
-        self.picked = self.picked % len(connections)
+
+        current_time_ms = datetime.utcnow()
+
+        while current_time_ms <= connections[self.picked]['time']:
```
ldmberman: We should only check each connection once here to avoid running late with respect to the user-specified timeout.

tests/conftest.py
```diff
@@ -172,11 +174,41 @@ def bdb_port():
...
+
 @fixture
 def bdb_node(bdb_host, bdb_port):
     return 'http://{host}:{port}'.format(host=bdb_host, port=bdb_port)
 
 
+@fixture
+def bdb_nodes_headers(bdb_node, bdb_node_header):
+    return [
+        {'endpoint': 'bigchaindb-driver_bigchaindb_2:9984',
```
ldmberman: Do the new tests require all these 5 instances to be up and running? How do you assert which endpoint was actually called?

tests/test_driver.py
```diff
@@ -151,6 +158,67 @@ def test_get(self, driver,
                        for tx in response if 'id' in tx['asset'])
 
 
+class TestOutputsEndpointMultiple:
```
ldmberman: The pecularities of the outputs', blocks', and other entities' structures are decoupled from how we do requests so I do not see much help in repeating all these `TestOutputsEndpointMultiple`  and `TestBlocksEndpointMultiple` tests.

I would take a single simple case, e. g. this one:

```
block = driver_multiple_headers.blocks.retrieve(block_height=str(block_with_alice_transaction))
assert block
```

and try to assert it works for different node setups.

bigchaindb_driver/transport.py
tucanae47: i agree so ill update accordingly on next push.

bigchaindb_driver/pool.py
tucanae47: I agree on the fact that it is actually failing! (ill fix it, need to think a test on that too i guess) is not respecting the user-specified timeout, but disagree on checking only once, more specifically:

1. We need to add the timeout check on that point too, (not figure out how yet).

2. But we could not stop the loop after calling only next nodes related to the connections length nor all of them once, as we don't know which node gets ready after its specific backoff time, couple of examples:

- When all nodes for a little while stop responding, and we need to find the very first one asap.

- When we are on hitting the last node, say ```ith= len(nodes)```, suddenly that one goes down and the next node waking up is ```ith-1```, we need to start hitting that one.
 
what do you think?

tests/test_driver.py
tucanae47: Ok ill update accordingly.  but the idea behind is not to select some specific request case for testing multiple nodes running, i selected a random couple of endpoint requests which could represent the complete set and add the multiple nodes test case there, maybe we should add them to all of them?  on the other hand yeah i agree if the representative sample to test multiple nodes is the one you suggest ill remove current ones and add only ```blocks.retrieve```

bigchaindb_driver/pool.py
ldmberman: Good point!

I see 2 approaches:

1. Update the time counter on every iteration of the loop in `forward_request`.
```
end = time()
time_left -= timedelta(seconds=end - start)
```
It is then fine to only check the nodes list once since it does not take a significant amount of time. If nothing is found the search will be simply repeated.

2. Pass the timeout (`utcnow() + time_left` in `forward_request`) to `pick` so that the while loop turns into `while current_time_ms <= connections[self.picked]['time'] and current_time_ms <= timeout`. Then you also need to alter `time_left` after the connection is found but before you make an HTTP request.

tests/conftest.py
ldmberman: > After that i stop-start 1 to 3 nodes nodes when the simulation (tests) is running to how see how the tests go.

This is great and definitely something worth doing! On the other hand, it is difficult to automate. So a better alternative for the automated test suite, like this one, is to mock the basic request method (e. g. `connection.request`) to behave differently depending on the URL passed, what allows to make all of the important assertions.

> main reason for the sleep(async wait)

`sleep` is an expensive approach to this problem. A better way is to retry "not found" requests - then you get closer to the time you actually need to spend to receive certain data. Off course, if you mock the request function, you do not have this problem at all.

tests/test_driver.py
ldmberman: Yes, all endpoints use the same transport so I do not see why would we want to query every endpoint with the new setup.

tucanae47:
Hey guys, i pushed new changes with last review suggestions. Main change besides review is related with @ldmberman suggestions regarding nodes normalization, so i (propose) _implicitly_ removed headers parameter from Transport class, and let normalization with headers to happen in utils, so we only deal with one kind of format to be handle internally which is:
```
[{'endpoint': 'https://test.bigchaindb.com',
     'headers': {'app_id': 'your_app_id',
                 'app_key': 'your_app_key'}},
    {'endpoint': 'https://test2.bigchaindb.com',
     'headers': {'app_id': 'your_app_id',
                 'app_key': 'your_app_key',
		 'extra_header': 'extra value'}}...]
```

ldmberman:
@tucanae47 thank you for the update! The only part I would like to further refine I described [here](https://github.com/bigchaindb/bigchaindb-driver/pull/430#issuecomment-397229028). What are your thoughts on it?

codegeschrei:
If you change something again, could you merge master into your branch again?
We had some issues with codecov and now it is working again :)

tucanae47:
@ldmberman  please have a look at the new push, basically im collecting a dict of errors raised per node:
```
exceptions = {}
while time_left.total_seconds() > 0:
    try:
        connection = self.get_connection(time_left)
        response = connection.request(..)
        return response.data
    except TransportError as err:
        end = time()
        time_left -= timedelta(seconds=end - start)
        self.pool.fail_node()
        start = time()
        exceptions[err.url] = {"code": err.status_code, "info": err.info}
    except BaseException as err:
      ...
exc_cls = HTTP_EXCEPTIONS.get(504, TransportError)
if len(exceptions):
    raise exc_cls(504, "Gateway Timeout", exceptions)
...
```
as of the actual error code and exception to return i think is good enough to use [aws http load balancers 504 error](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/ts-elb-error-message.html#ts-elb-errorcodes-http504)

bigchaindb_driver/transport.py
```diff
@@ -74,14 +74,23 @@ def forward_request(self, method, path=None,
                     headers=headers
                 )
                 return response.data
+            except TransportError as err:
+                end = time()
+                time_left -= timedelta(seconds=end - start)
+                self.pool.fail_node()
+                start = time()
+                code = None if err.status_code is None else err.status_code
```
ldmberman: You can just use `err.status_code`.

bigchaindb_driver/transport.py
```diff
@@ -74,14 +74,23 @@ def forward_request(self, method, path=None,
...
+                info = None if err.info is None else err.info
+                exceptions[err.url] = {"code": code, "info": info}
```
ldmberman: I would collect them into a list - currently we can lose some information if the same node experiences 2 different exceptions. What do you think about a list of pairs `(url, error)`?

ldmberman:
> i think is good enough to use aws http load balancers 504 timeout error

It's appropriate if you implement an HTTP server. Here you might make the user think there is actually an HTTP server somewhere which closed the connection after a timeout.

What do you think about defining a custom exception and raising it there?
```
class TimeoutException(BigchaindbException):
   pass
```

bigchaindb_driver/transport.py
tucanae47: if we do list, we are going to repeat errors, as we would easily have the same error multiple times and i dont think we will lose relevant info, we are giving the last error why every node failed.

ldmberman: Repeated errors just reflect the real history of requests and might provide useful hints. Repeating is not always bad.

The current implementation should also work just fine - you can keep it.

tucanae47: Ok, updated with pair instead of dict, and keep the dict. Other suggestions added on next push.

tucanae47:
Hi, just merged master and it seems codecov is failing, not sure how to fix it.

codegeschrei:
Hey, the codecov can be ignored for now. It is just complaining about one line :)

PR #435, opened 2018-06-01
Problem: The README is not userfriendly enough

codegeschrei:
Solution: update the README and add an usage example

removed some links that are not used anymore
added a TOC
added a user example
removed pyup badge
added some helpful links


manolodewiner:
Nice PR, makes sense to add examples.
Also the JS driver has an example on how to use the driver

PR #437, opened 2018-06-04
Documentation fix

### LINKED ISSUE #436: Basic Usage Examples incorrect documentation
> Also note how, because we were consuming a ``TRANSFER`` transaction, we were
> able to directly use the ``TRANSFER`` transaction's ``asset`` as the new
> transaction's ``asset`` because it already contained the asset's id.

Actually if using the example, a ``CREATE`` transaction is consumed, and its ID is used.

aruseni:
Actually if using the example, a ``CREATE`` transaction is consumed, and its ID is used.

Fixes #436

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -498,10 +498,6 @@ ensures that the amount being consumed in each transaction (with divisible
 assets) is the same as the amount being output. This ensures that no amounts
 are lost.
 
-Also note how, because we were consuming a ``TRANSFER`` transaction, we were
-able to directly use the ``TRANSFER`` transaction's ``asset`` as the new
-transaction's ``asset`` because it already contained the asset's id.
-
 The ``fulfilled_transfer_tx`` dictionary should have two outputs, one with
 ``amount='2'`` and the other with ``amount='8'``:
 


docs/usage.rst
```diff
@@ -498,10 +498,6 @@ ensures that the amount being consumed in each transaction (with divisible
 assets) is the same as the amount being output. This ensures that no amounts
 are lost.
 
-Also note how, because we were consuming a ``TRANSFER`` transaction, we were
-able to directly use the ``TRANSFER`` transaction's ``asset`` as the new
-transaction's ``asset`` because it already contained the asset's id.
-
```
codegeschrei: instead of removing it, can you just change the parts that are wrong?

aruseni:
@codegeschrei Take a look at this file. There are 2 places where it is mentioned that the ID of the asset is the ID of the transaction which was used for creating the asset:

* In **Asset Transfer**
* In **Querying for Transactions**

So do you think it is necessary to mention it another time? IMHO keeping things short is always a good idea if it does not interfere with the ease to understand the documentation.

ttmc:
> ...keeping things short is always a good idea if it does not interfere with the ease to understand the documentation.

I suppose that's a reasonable reason to remove it.

PR #440, opened 2018-06-05
Problem: the default mode of a tx does not match the default of the s…

### LINKED ISSUE #431: Changes to the method(s) for posting/sending a transaction
This issue is to implement what was discussed in https://github.com/bigchaindb/bigchaindb/issues/2307

For details, see the comment https://github.com/bigchaindb/bigchaindb/issues/2307#issuecomment-391964164

codegeschrei:
…erver

Solution: remove the default mode and add a method for each mode

- added deprecation warning for `send`
- added a method for each mode
- updated tests and docs
- changed one fixture and added another to have more unique transactions. Otherwise I would have not been able to run the tests for the mode (yay)

(darn it, my Problem statement is too long)

Resolves #431

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,7 +99,7 @@ Afterwards Bob spends 3 of these tokens.
     fulfilled_token_tx = bdb.transactions.fulfill(
         prepared_token_tx,
         private_keys=alice.private_key)
-    bdb.transactions.send(fulfilled_token_tx, mode='commit')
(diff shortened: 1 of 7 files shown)


ttmc:
@z-bowen The options were discussed in https://github.com/bigchaindb/bigchaindb/issues/2307 . Having one `send()` method with a default mode (commit) was how it worked before.

PR #443, opened 2018-06-13
prepare release

codegeschrei:


ttmc:
There's an error in Travis:
```text
    def test_info(self, driver):
        response = driver.info()
        assert 'api' in response
        assert 'docs' in response
>       assert response['keyring'] == []
E       KeyError: 'keyring'
```
The "info" endpoint doesn't return the "keyring" or "keypair" any more, now that PR bigchaindb/bigchaindb#2338 is merged. I'll see if I can push a fix.

ttmc:
To do first in separate PRs before merging this PR:
- Update the docs
- Check the method that calls the /blocks endpoint, to see if it's still working or needs to be updated. (The [BigchainDB HTTP API still has a /blocks endpoint](https://docs.bigchaindb.com/projects/server/en/latest/http-client-server-api.html#blocks), but its syntax and response have changed.)

PR #444, opened 2018-06-14
Problem: info endpoint does not return keyrings

codegeschrei:
Solution: remove keyrings from tests and docs


bigchaindb_driver/driver.py
```diff
@@ -101,8 +101,6 @@ def info(self, headers=None):
 
                 * the server version,
                 * the public key of the node, and
```
ttmc: The public key of the node isn't returned any more either.

PR #450, opened 2018-06-18
 Problem: the demo script in README does not work 

### LINKED ISSUE #449: Demo script: 'bike_token_id' is not defined
* bigchaindb-driver version: v0.5.0
* bigchaindb **server** version (`bigchaindb -v`): 2.0.0b1
* Python version: 3.5.2
* Operating System: Ubuntu

### Description

Demo from the driver's landing page fails with an error about 'bike_token_id' being undefined.

### What I Did

ran the demo script
```
nameError: name 'bike_token_id' is not defined
```

codegeschrei:
Solution: use the correct tx id with transfer step

## Issues This PR Fixes
Fixes #449

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -103,7 +103,7 @@ Afterwards Bob spends 3 of these tokens.
 
     # Use the tokens
     # create the output and inout for the transaction
-    transfer_asset = {'id': bike_token_id}
+    transfer_asset = {'id': fulfilled_token_tx['id']}
     output_index = 0
     output = fulfilled_token_tx['outputs'][output_index]
     transfer_input = {'fulfillment': output['condition']['details'],
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,8 @@ Usage
 Example: Create a divisible asset for Alice who issues 10 token to Bob so that he can use her Game Boy.
 Afterwards Bob spends 3 of these tokens.
 
+If you want to send a transaction you need to `Determine the BigchainDB Root URL`_.
+
 .. code-block:: python
 
     # import BigchainDB and create an object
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -103,13 +103,12 @@ Afterwards Bob spends 3 of these tokens.
...
     # create the output and inout for the transaction
-    transfer_asset = {'id': bike_token_id}
+    transfer_asset = {'id': fulfilled_token_tx['id']}
     output_index = 0
     output = fulfilled_token_tx['outputs'][output_index]
     transfer_input = {'fulfillment': output['condition']['details'],
                       'fulfills': {'output_index': output_index,
-                                   'transaction_id': fulfilled_token_tx[
-                                       'id']},
+                                   'transaction_id': transfer_asset},
```
ldmberman: It should be `transfer_asset['id']`.

ldmberman:
What do you think about defining `bdb_root_url = 'https://example.com:9984'` via `localhost`? - people would be able to try the script out locally, without any modifications.

codegeschrei:
I see your point but not all of them use it locally. Some want to use it with the test network, so I think something generic that tells you to change it, is a good starting point.

README.rst
```diff
@@ -69,15 +69,15 @@ Afterwards Bob spends 3 of these tokens.
 
     # import BigchainDB and create an object
     from bigchaindb_driver import BigchainDB
-    bdb_root_url = 'https://example.com:9984'
+    bdb_root_url = 'http://localhost:32804'
```
ldmberman: Why did you change the port to `32804`?

codegeschrei: ah, shoot.. I reverted it twice and it got back to my test settings. I'll change it to the original

ldmberman:
What do you think about both using `localhost` and adding an additional note about it?

codegeschrei:
I can put a link to the [docs](https://docs.bigchaindb.com/projects/py-driver/en/latest/connect.html)

PR #456, opened 2018-06-21
Problem: The round robin strategy has issues.

ldmberman:
Solution: Improve the implementation:
- remove busy wait;
- only retry connection errors;
- simplify the code;
- make custom headers overwrite common headers, not vice versa;
- add some more tests;
- remove obsolete await_transaction;
- make the linter pass.

### DIFF
--- a/bigchaindb_driver/connection.py
+++ b/bigchaindb_driver/connection.py
@@ -36,11 +36,22 @@ class Connection:
 
     def request(self, method, *, path=None, json=None,
                 params=None, headers=None, timeout=None, **kwargs):
-        """Performs an HTTP requests for the specified arguments.
+        """Performs an HTTP request with the given parameters.
 
-           If the node has backoff time attached, waits till the time is due.
+           Implements exponential backoff.
 
-           Backoff time is adjusted after the request is made.
+           If `ConnectionError` occurs, a timestamp equal to now +
+           the default delay (`BACKOFF_DELAY`) is assigned to the object.
+           The timestamp is in UTC. Next time the function is called, it either
+           waits till the timestamp is passed or raises `TimeoutError`.
+
+           If `ConnectionError` occurs two or more times in a row,
+           the retry count is incremented and the new timestamp is calculated
+           as now + the default delay multiplied by two to the power of the
+           number of retries.
+
+           If a request is successful, the backoff timestamp is removed,
+           the retry count is back to zero.
 
         Args:
             method (str): HTTP method (e.g.: ``'GET'``).
--- a/bigchaindb_driver/connection.py
+++ b/bigchaindb_driver/connection.py
@@ -1,15 +1,21 @@
+import time
+
 from collections import namedtuple
+from datetime import datetime, timedelta
 
 from requests import Session
+from requests.exceptions import ConnectionError
 
 from .exceptions import HTTP_EXCEPTIONS, TransportError
 
+
+BACKOFF_DELAY = 0.5  # seconds
 
 HttpResponse = namedtuple('HttpResponse', ('status_code', 'headers', 'data'))
 
 
 class Connection:
-    """A Connection object to make HTTP requests."""
+    """A Connection object to make HTTP requests to a particular node."""
 
     def __init__(self, *, node_url, headers=None):
         """Initializes a :class:`~bigchaindb_driver.connection.Connection`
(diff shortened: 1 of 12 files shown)


bigchaindb_driver/driver.py
```diff
@@ -1,40 +1,40 @@
...
-                (e.g. :meth:`BigchainDB().transactions.send_commit()
-                <.TransactionsEndpoint.send_commit>`).
-
+           Args:
+             *nodes (list of (str or dict)): BigchainDB nodes to connect to.
+               Currently, the full URL must be given. In the absence of any
+               node, the default(``'http://localhost:9984'``) will be used.
+               If node is passed as a dict, `endpoint` is a required key;
+              `headers` is an optional `dict` of headers.
+               transport_class: Optional transport class to use.
```
codegeschrei: the indentation for `transport_class` is wrong

bigchaindb_driver/transport.py
```diff
@@ -1,50 +1,50 @@
...
 
 
 class Transport:
-    """Transport class."""
+    """Transport class.
+
+    """
 
-    def __init__(self, *nodes, timeout=None, headers=None):
+    def __init__(self, *nodes, headers=None, timeout=None):
```
codegeschrei: are we using the `headers` here somewhere?

ldmberman: Great catch! We do not need them here.

codegeschrei:
Would it make sense to have a test case for the `TimeoutError`?

bigchaindb_driver/connection.py
```diff
@@ -25,36 +31,79 @@ def __init__(self, *, node_url, headers=None):
...
+                **kwargs,
+            )
+        except ConnectionError as err:
+            connExc = err
+            raise err
+        finally:
+            self.update_backoff_time(success=connExc is None)
+        return response
+
+    def get_backoff_timedelta(self, timeout=None):
```
codegeschrei: We don't really use the `timeout` in here

ldmberman: Good catch!

bigchaindb_driver/connection.py
```diff
@@ -25,36 +31,79 @@ def __init__(self, *, node_url, headers=None):
...
+        self._retries = 0
+        self.backoff_time = None
+
+    def request(self, method, *, path=None, json=None,
+                params=None, headers=None, timeout=None, **kwargs):
         """Performs an HTTP requests for the specified arguments.
 
+           If the node has backoff time attached, waits till the time is due.
+
+           Backoff time is adjusted after the request is made.
```
codegeschrei: The description here could be a bit more detailed. When I just read it like that I ask myself where the backoff time is coming from and what it is. Next to that `backoff_time` is not directly used in here, that confused me in the beginning.

codegeschrei: Can you go a bit more into detail with the comment here? In the beginning it was a bit confusing for me. I also didn't get why there was no `backoff_time` in the function.

codegeschrei: sorry for commenting twice, I thought I forgot to do it last week :D

codegeschrei:
I'm wondering if we should add documentation for it. Maybe a small example how to add nodes and to understand when which error can occur?

bigchaindb_driver/transport.py
```diff
@@ -1,50 +1,48 @@
...
         """Initializes an instance of
         :class:`~bigchaindb_driver.transport.Transport`.
 
         Args:
-            nodes: nodes
-            timeout: timeout
-            headers (dict): Optional headers to pass to the
-                :class:`~.connection.Connection` instances, which will
-                add it to the headers to be sent with each request.
+            nodes: each node is a dictionaty with the keys `endpoint` and
```
codegeschrei: There is a typo in `dictionary`

bigchaindb_driver/driver.py
```diff
@@ -1,40 +1,40 @@
...
-                (e.g. :meth:`BigchainDB().transactions.send_commit()
-                <.TransactionsEndpoint.send_commit>`).
-
+           Args:
+             *nodes (list of (str or dict)): BigchainDB nodes to connect to.
+               Currently, the full URL must be given. In the absence of any
+               node, the default(``'http://localhost:9984'``) will be used.
+               If node is passed as a dict, `endpoint` is a required key;
+               `headers` is an optional `dict` of headers.
+               transport_class: Optional transport class to use.
```
codegeschrei: the indentation needs to be fixed here.

ldmberman: Can you elaborate? I do not notice any issues.

codegeschrei: before the indentation of the `transport_class` was different https://github.com/bigchaindb/bigchaindb-driver/pull/456/files#diff-0de47fef8ba327e59600c5d4d879e81eL27
It's to separate all the args

bigchaindb_driver/connection.py
```diff
@@ -25,36 +31,90 @@ def __init__(self, *, node_url, headers=None):
...
-        )
+        backoff_timedelta = self.get_backoff_timedelta()
+
+        if timeout is not None and timeout < backoff_timedelta:
+            raise TimeoutError
+
+        if backoff_timedelta > 0:
+            time.sleep(backoff_timedelta)
+
+        connExc = None
```
kansi: what about `connExc = True` ? then we don't need `connExc is None`

kansi: I didn't get you are trying to imply here. What I was trying to say is that if `connExc` is `None` is always converted to `True` we could rather just assign it a default value of `True`. You can choose either way

ldmberman: `connExc` represents an exception - that's why I do not want to give it a default value of `True` since `True` is not really a value of an exception.

PR #459, opened 2018-07-16
Problem: bdb voting does not exist anymore

codegeschrei:
Solutions: remove leftover exceptions


codegeschrei:
in bdb we updated tendermint. I need to include the changes in master first and merge them in here to let the tests run

PR #466, opened 2018-08-14
WIP - Problem: common is outdated

codegeschrei:
Solution: add changes according to bdb common

I think what is left is to fix the tests


codegeschrei:
All changes in common are related to the `upsert-validator` which is not needed for the driver. Closing this for now.

PR #471, opened 2018-08-28
Problem: timeout is not passed to backoff capper.

ldmberman:
Solution: pass timeout to backoff capper so that the cap depends on the timeout.


ldmberman:
The timeout (half of it) actually needs to be static otherwise we approach the very frequent retries again - I will fix it.

bigchaindb_driver/connection.py
```diff
@@ -40,7 +39,8 @@ def __init__(self, *, node_url, headers=None):
         self.backoff_time = None
 
     def request(self, method, *, path=None, json=None,
-                params=None, headers=None, timeout=None, **kwargs):
+                params=None, headers=None, timeout=None,
+                backoff_cap=None, **kwargs):
```
codegeschrei: can you add the `backoff_cap` to the docstring?

Issue #475, opened 2018-09-03
Problem: All recent builds of the Python driver docs are failing on ReadTheDocs


ttmc:
Here's a list of all recent builds of the Python Driver docs. They are all failing. From https://readthedocs.org/projects/bigchaindb-python-driver/builds/

![screenshot from 2018-09-03 16-54-35](https://user-images.githubusercontent.com/1210951/44993210-1364d600-af9a-11e8-99f1-543e11859e22.png)

When I build the Python Driver docs locally, the build works, but there are many exceptions raised, and they seem to be related to the recent changes:

```bash
(drv) troy@troy-thinkpad:~/repos/bigchaindb-driver/docs$ make html
sphinx-build -b html -d _build/doctrees   . _build/html
Running Sphinx v1.7.6
making output directory...
loading pickled environment... not yet created
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://docs.bigchaindb.com/projects/server/en/latest/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 13 source files that are out of date
updating environment: 13 added, 0 changed, 0 removed
reading sources... [ 61%] handcraft                                                                                                    

>>>-------------------------------------------------------------------------
Exception in /home/troy/repos/bigchaindb-driver/docs/handcraft.rst at block ending on line 423
Specify :okexcept: as an option in the ipython:: block to suppress this message
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-43-fe4a338e75cc> in <module>()
----> 1 json_str_handcrafted_tx = json.dumps(handcrafted_creation_tx, sort_keys=True)

/usr/lib/python3.6/json/__init__.py in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    236         check_circular=check_circular, allow_nan=allow_nan, indent=indent,
    237         separators=separators, default=default, sort_keys=sort_keys,
--> 238         **kw).encode(obj)
    239 
    240 

/usr/lib/python3.6/json/encoder.py in encode(self, o)
    197         # exceptions aren't as detailed.  The list call should be roughly
    198         # equivalent to the PySequence_Fast that ''.join() would do.
--> 199         chunks = self.iterencode(o, _one_shot=True)
    200         if not isinstance(chunks, (list, tuple)):
    201             chunks = list(chunks)

[...]

PR #476, opened 2018-09-06
Problem: Error when building Handcrafting Transactions docs

ttmc:
The docs would build, but the generated "Handcrafting Transactions" page was full of valid Python errors of the form:
`TypeError: Object of type 'bytes' is not JSON serializable`

I guess sometimes people _want_ to show the actual error messages in their docs, so such errors don't prevent Sphinx from thinking everything is okay.

Solution: Use `base58.b58encode(key).decode()` which is a str, instead of `base58.b58encode(key)` which is a bytes object. I defined a little helper function named `b58()` to reduce verbosity.

I don't know if this will get ReadTheDocs to start working (as per issue #475), but it does remove all the "Traceback" messages in the "Handcrafting Transactions" page of the docs.

### DIFF
--- a/docs/handcraft.rst
+++ b/docs/handcraft.rst
@@ -273,9 +273,13 @@ As for the details:
 
 .. ipython::
 
+    In [0]: def b58(key):
+       ...:     # convert key to a base58 string
+       ...:     return base58.b58encode(key).decode()
+
     In [0]: condition_details = {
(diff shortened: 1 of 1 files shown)


ttmc:
@codegeschrei I'm not sure why some tests in Travis are failing. They seem unrelated to the changes I made in this pull request.

PR #478, opened 2018-09-10
Problem: the tests are failing

codegeschrei:
Solution: rewrite test without fixtures

I couldn't find the underlying problem with the fixtures. In general we had difficulties in the past. Fixtures being reused and the tests being not independent led to errors and failing tests. It would be nice to find the reason why they are failing but worst case I waste a lot of time on it and I'm already out of ideas how to debug it.

I tried to make the test as short as possible, if you see any lines I can shorten, please tell me.

### DIFF
--- a/tests/test_driver.py
+++ b/tests/test_driver.py
@@ -185,37 +185,73 @@ class TestOutputsEndpoint:
             'output_index': 0
         } in outputs
 
-    @mark.parametrize('spent,outputs_qty', ((False, 1), (True, 1), (None, 2)))
-    def test_get_outputs_with_spent_query_param(
-            self, spent, outputs_qty, driver, carol_pubkey,
-            persisted_carol_bicycle_transaction,
-            persisted_carol_car_transaction,
-            persisted_transfer_carol_car_to_dimi):
-        outputs = driver.outputs.get(carol_pubkey, spent=spent)
-        assert len(outputs) == outputs_qty
+    def test_get_outputs_with_spent_query_param(self, driver):
+        from bigchaindb_driver.crypto import generate_keypair
+        import uuid
 
-        # Return only unspent outputs
-        if spent is False:
-            assert {
-                'transaction_id': persisted_carol_bicycle_transaction['id'],
-                'output_index': 0
-            } in outputs
-        # Return only spent outputs
-        elif spent is True:
-            assert {
-                'transaction_id': persisted_carol_car_transaction['id'],
-                'output_index': 0
-            } in outputs
-        # Return all outputs for carol
-        elif spent is None:
-            assert {
-                'transaction_id': persisted_carol_bicycle_transaction['id'],
-                'output_index': 0
-            } in outputs
-            assert {
-                'transaction_id': persisted_carol_car_transaction['id'],
-                'output_index': 0
-            } in outputs
+        def create_transaction():
+            return driver.transactions.prepare(
+                    operation='CREATE',
+                    signers=carol.public_key,
+                    asset={
+                        'data': {
+                            'asset': {
+                                'serial_number': str(uuid.uuid4()),
+                                'manufacturer': str(uuid.uuid4()),
+                            },
+                        },
+                    },
+            )
+
+        carol, dimi = generate_keypair(), generate_keypair()
+
+        assert len(driver.outputs.get(carol.public_key, spent=True)) == 0
+        assert len(driver.outputs.get(carol.public_key, spent=False)) == 0
+        assert len(driver.outputs.get(carol.public_key, spent=None)) == 0
+
+        # create the first transaction for carol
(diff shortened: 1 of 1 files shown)


tests/test_driver.py
```diff
@@ -185,37 +185,69 @@ def test_get_outputs(self, driver, carol_pubkey,
...
-                'transaction_id': persisted_carol_car_transaction['id'],
-                'output_index': 0
-            } in outputs
+    def test_get_outputs_with_spent_query_param(self, driver):
+        from bigchaindb_driver.crypto import generate_keypair
+        import uuid
+
+        outputs = [(False, 2, 1), (True, 0, 1), (None, 2, 2)]
+
+        def create_asset():
```
ttmc: The name of this function (`create_asset()`) is a bit misleading because it prepares a CREATE transaction, it doesn't really create an asset. To create an asset, I think you'd also have to fulfill the CREATE transaction, post the result to a BigchainDB network, and get it committed to the blockchain. A name like `prepare_create_tx()` would be more accurate.

codegeschrei: ah, ja.. I named it when it was still just creating an asset and forgot to rename it later. I'll change that :)

tests/test_driver.py
```diff
@@ -185,37 +185,69 @@ def test_get_outputs(self, driver, carol_pubkey,
...
-            } in outputs
-            assert {
-                'transaction_id': persisted_carol_car_transaction['id'],
-                'output_index': 0
-            } in outputs
+    def test_get_outputs_with_spent_query_param(self, driver):
+        from bigchaindb_driver.crypto import generate_keypair
+        import uuid
+
+        outputs = [(False, 2, 1), (True, 0, 1), (None, 2, 2)]
```
ttmc: These aren't "outputs" in the BigchainDB sense of the word "outputs." (A BigchainDB output is characterized by a transaction ID and an output index.) These things are just test vectors or something like that.

ttmc: That didn't bother me at first, but this test is testing "outputs" in the BigchainDB sense of the word, so the semantics got quite confusing after a while.

codegeschrei: Changed it :)

tests/test_driver.py
```diff
@@ -185,37 +185,69 @@ def test_get_outputs(self, driver, carol_pubkey,
...
+        create_tx1 = create_transaction()
+        create_tx1 = driver.transactions.fulfill(
+            create_tx1, private_keys=carol.private_key)
+        driver.transactions.send_commit(create_tx1)
+        # create the second transaction for carol
+        create_tx2 = create_transaction()
+        create_tx2 = driver.transactions.fulfill(
+            create_tx2, private_keys=carol.private_key)
+        driver.transactions.send_commit(create_tx2)
+        for entry in test_parameters:
```
vrde: I think it's easier to read if you just unwrap this loop and write the three `assert`s. You can also remove `test_parameters` and just use the values directly.

tests/test_driver.py
```diff
@@ -185,37 +185,69 @@ def test_get_outputs(self, driver, carol_pubkey,
...
+                        'data': {
+                            'asset': {
+                                'serial_number': str(uuid.uuid4()),
+                                'manufacturer': str(uuid.uuid4()),
+                            },
+                        },
+                    },
+            )
+
+        carol, dimi = generate_keypair(), generate_keypair()
```
vrde: It might sound dumb, but I would add the initial conditions of this test, that are:

```python
assert len(driver.outputs.get(carol.public_key, spent=True)) == 0
assert len(driver.outputs.get(carol.public_key, spent=False)) == 0
assert len(driver.outputs.get(carol.public_key, spent=None)) == 0
```

tests/test_driver.py
```diff
@@ -185,37 +185,69 @@ def test_get_outputs(self, driver, carol_pubkey,
...
+            asset=transfer_asset,
+            inputs=transfer_input,
+            recipients=dimi.public_key,
+        )
+        transfer_tx = driver.transactions.fulfill(
+            transfer_tx, private_keys=carol.private_key,
+        )
+        driver.transactions.send_commit(transfer_tx)
+
+        for entry in test_parameters:
```
vrde: Here I have the same comment I made above about unwrapping the loop.

tests/test_driver.py
codegeschrei: I'll change it but do you have a specific reason?

vrde: In this test we check that the `outputs` endpoint works as expected. To me, another condition to check is that there are no outputs for a freshly generated keypair, something like "when I start, the outputs are zero, then I add an asset and I expect 1, then ..."

codegeschrei: Yes, that makes sense!

PR #483, opened 2018-09-17
Problem: The send function is deprecated

### LINKED ISSUE #482: send() is deprecated, remove it for the next release
With the release of version `0.5.0` we deprecated the `send` and instead `send_commit()`, `send_async()`, and `send_sync()` needs to be used. 

This needs to be done now, since we didn't remove it in the last two releases

codegeschrei:
Solution: remove the send function

Fixes #482

### DIFF
--- a/bigchaindb_driver/driver.py
+++ b/bigchaindb_driver/driver.py
@@ -5,7 +5,6 @@
 from .transport import Transport
 from .offchain import prepare_transaction, fulfill_transaction
 from .utils import normalize_nodes
-from warnings import warn
 
 
 class BigchainDB:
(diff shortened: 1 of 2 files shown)


bigchaindb_driver/driver.py
```diff
@@ -318,41 +317,6 @@ def get(self, *, asset_id, operation=None, headers=None):
...
-
-        .. warning::
-
-            The method .send will be deprecated in the next release
-            of the driver, please use ``.send_commit``, ``.send_sync``, or
-            ``.send_async`` instead. More info:
-            https://docs.bigchaindb.com/projects/py-driver/en/latest/handcraft.html#send-the-transaction
-        """
-
-        warn('The method .send will be deprecated in the next release of the '
```
ttmc: Does this mean that all the docs warning messages, like this

screenshot from 2018-09-17 15-10-09

will go away, or do you have to remove those manually?

ttmc: Neat.

